PR #47, opened 2025-11-27
fix: restrict proxy paths that can be forwarded to cloud-api

### LINKED ISSUE #23: Potential Security Risk: Unrestricted URL Forwarding
Currently, chat-api forwards requests without imposing restrictions on the target URLs. This behavior could lead to unintended misuse of the configured cloud-api API key. Specifically, if a user or external actor forwards requests to arbitrary endpoints, the API key could be exposed or used in ways not intended by the system.

hanakannzashi:
Fixes #23

### DIFF
--- a/crates/api/src/consts.rs
+++ b/crates/api/src/consts.rs
@@ -2,4 +2,10 @@ pub const SYSTEM_PROMPT_MAX_LEN: usize = 64 * 1024;
 pub const LIMIT_MAX: i64 = 100;
 
 /// Allowed proxy paths that can be forwarded to OpenAI
-pub const ALLOWED_PROXY_PATHS: &[&str] = &["responses", "chat/completions", "models"];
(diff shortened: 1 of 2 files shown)


crates/api/src/consts.rs
```diff
@@ -1,2 +1,5 @@
 pub const SYSTEM_PROMPT_MAX_LEN: usize = 64 * 1024;
 pub const LIMIT_MAX: i64 = 100;
+
+/// Allowed proxy paths that can be forwarded to OpenAI
+pub const ALLOWED_PROXY_PATHS: &[&str] = &["responses", "chat/completions", "models"];
```
think-in-universe: Let's carefully review the endpoints that we'd like to whitelist. I think we're missing below endpoints:

- `/model`: e.g. `/model/list`
- `/conversations`
- `/files`
- `/signature`

think-in-universe (changes requested):

PR #51, opened 2025-11-28
feat: files api

### LINKED ISSUE #49: Security issue: /files Endpoint Exposes Other Users' Files
### Description
The /files endpoint currently allows users to access files uploaded by other users. This occurs because requests are directly forwarded to cloud-api without proper access control checks.

---
Observed Behavior

- Any authenticated user can query /files and receive file metadata or content belonging to other users.
- There is no authorization or ownership validation before forwarding the request to cloud-api.

---
### Expected Behavior

- Users should only be able to access their own files.
- Requests should include proper permission checks before fetching data from the backend or forwarding to cloud-api.
- Unauthorized access must be blocked and return a 403 Forbidden error.

hanakannzashi:
Fix #49

### DIFF
--- a/crates/database/src/migrations/sql/V8__add_files.sql
+++ b/crates/database/src/migrations/sql/V8__add_files.sql
@@ -13,7 +13,6 @@ CREATE TABLE files (
 
 CREATE INDEX idx_files_user_id ON files(user_id);
 CREATE INDEX idx_files_file_created_at ON files(file_created_at DESC);
-CREATE INDEX idx_files_updated_at ON files(updated_at DESC);
 CREATE INDEX idx_files_purpose ON files(purpose);
 
 -- Trigger for files table


crates/database/src/migrations/sql/V8__add_files.sql
```diff
@@ -0,0 +1,18 @@
+-- Files table to track OpenAI files per user
+CREATE TABLE files (
+    id VARCHAR(255) PRIMARY KEY, -- OpenAI file ID
+    user_id UUID NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
```
think-in-universe: Should we track file names as well so it's easier to list and browse all the files owned by the user?

PR #84, opened 2025-12-18
feat: near account balance check

hanakannzashi:


crates/config/src/lib.rs
```diff
@@ -242,6 +243,22 @@ impl TelemetryConfig {
     }
 }
 
+/// NEAR-related configuration (shared between services)
+#[derive(Debug, Clone, Deserialize)]
+pub struct NearConfig {
+    /// NEAR JSON-RPC endpoint used for on-chain queries (e.g. balance checks)
+    pub rpc_url: Url,
```
think-in-universe: We can use the same NEAR RPC URL for NEAR wallet login

crates/api/src/routes/api.rs
```diff
@@ -1137,6 +1145,120 @@ async fn proxy_responses(
...
+
+    tracing::info!(
+        "NEAR balance for account '{}' (user_id={}): {} yoctoNEAR",
+        account_id_str,
+        user.user_id,
+        balance
+    );
+
+    if balance < MIN_NEAR_BALANCE {
+        tracing::warn!(
```
think-in-universe: Can we logout the user and limit the account's usage for a while?

think-in-universe (changes requested):

crates/api/src/routes/oauth.rs
```diff
@@ -478,6 +480,63 @@ pub async fn mock_login(
...
+                    user.id
+                );
+                None
+            }
+        };
+
+        if let Some(provider) = provider {
+            // Derive provider_user_id from email.
+            // For NEAR we expect email like `alice.near@near` and use `alice.near` as provider_user_id.
+            let provider_user_id = match provider {
```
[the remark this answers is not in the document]
hanakannzashi: This a mock API, follow the previous NEAR login logic

crates/api/src/routes/api.rs
```diff
@@ -1052,6 +1064,12 @@ async fn proxy_responses(
         user.session_id
     );
 
+    // Check if user is currently banned before proceeding
+    ensure_user_not_banned(&state, &user).await?;
+
+    // If user has a NEAR-linked account, enforce minimum balance before proxying
+    ensure_near_balance_for_near_user(&state, &user).await?;
```
think-in-universe: Let's check balance asynchronously.

env.example
```diff
@@ -68,5 +68,9 @@ CORS_ALLOWED_ORIGINS=http://localhost:3000
 # Expected recipient for signed messages (defaults to private.near.ai)
 NEAR_EXPECTED_RECIPIENT=private.near.ai
 
+# NEAR Configuration
+# Required: NEAR JSON-RPC endpoint used for on-chain queries (e.g. balance checks)
+NEAR_RPC_URL=https://near.lava.build
```
think-in-universe:
```suggestion
NEAR_RPC_URL=https://free.rpc.fastnear.com
```

think-in-universe (changes requested):

PR #92, opened 2025-12-26
fix: disable refreshing cloud-api access token

### LINKED ISSUE #79: Bug: race condition of refreshing cloud-api access token
## Problem

Currently chat-api authenticates with the cloud-api access token to set and get system prompt of its organization in cloud-api, which leads to race condition when chat-api is refreshing access token and may cause some unexpected requests failure: https://github.com/nearai/chat-api/blob/b16fad3ac9aa8f61c14cef438c12e6d214b6eb18/crates/services/src/vpc/service.rs#L257-L286

The logs below show that refreshing access token failed and is re-authenticated at 09:11:55, while leads to the batch conversation API call failure at 09:11:56.

```log
"2025-12-15T09:11:55.302Z","cpu01-chat-api-staging","chat-api","{""level"":""WARN"",""fields"":{""message"":""Failed to refresh access token, will re-authenticate: Refresh token expired""},""timestamp"":""2025-12-15T09:11:55.302552Z"",""target"":""services::vpc::service""}"

"2025-12-15T09:11:55.302Z","cpu01-chat-api-staging","chat-api","{""level"":""INFO"",""fields"":{""message"":""Performing full VPC authentication...""},""timestamp"":""2025-12-15T09:11:55.302578Z"",""target"":""services::vpc::service""}"

"2025-12-15T09:11:56.812Z","cpu01-chat-api-staging","chat-api","{""level"":""ERROR"",""fields"":{""message"":""Failed to list conversations for user_id=be20359b-baf6-43ad-8e1b-b2754a543206: OpenAI API error: OpenAI batch API returned status 401""},""timestamp"":""2025-12-15T09:11:56.812944Z"",""target"":""api::routes::api""}"
```

#### Simple Workaround

Add default retry for all `forward_request` if failed due to 401 error.

#### Complete Fix

Get rid of access token completely in chat-api. Access token is supposed to be used by frontend. API key should be used by the backend service like chat-api. Now the access token in chat-api is only used for getting and setting system prompt in cloud-api. 

We have two ways to fix this:

1. Allow API key authentication for organization settings in cloud-api
2. Implement the system prompts settings in model settings of chat-api (Reference #74).

hanakannzashi:
Fixes #79

### DIFF
--- a/crates/services/src/vpc/ports.rs
+++ b/crates/services/src/vpc/ports.rs
@@ -31,7 +31,7 @@ pub struct VpcAuthConfig {
 /// Service for managing VPC credentials
 #[async_trait]
 pub trait VpcCredentialsService: Send + Sync {
-    /// Get VPC credentials, requesting new tokens if necessary
+    /// Get VPC credentials, re-authenticating and requesting new credentials if necessary
     /// Returns None if VPC is not configured
     async fn get_credentials(&self) -> anyhow::Result<Option<VpcCredentials>>;
 
--- a/crates/services/src/vpc/service.rs
+++ b/crates/services/src/vpc/service.rs
@@ -32,6 +32,7 @@ struct VpcOrganization {
     id: String,
 }
 
+/// In-memory cached VPC authentication data.
 struct CachedCredentials {
     organization_id: String,
     api_key: String,


think-in-universe:
@claude review

PR #113, opened 2026-01-13
feat: user rate limit with time window

hanakannzashi:


crates/services/src/analytics/ports.rs
```diff
@@ -187,3 +231,32 @@ pub enum AnalyticsError {
     #[error("Internal error: {0}")]
     InternalError(String),
 }
+
+/// Time window for sliding window rate limiting (in days)
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct TimeWindow {
+    /// Number of days for the sliding window
+    pub days: u32,
```
think-in-universe: Should we also allow minutes and hours?

crates/api/src/middleware/rate_limit.rs
```diff
@@ -356,4 +634,1085 @@ mod tests {
...
+            assert_eq!(reported_limit, limit);
+            // retry_after_ms should be approximately 1 day in milliseconds
+            let expected_retry_ms = 24 * 3600 * 1000;
+            assert_eq!(retry_after_ms, expected_retry_ms);
+        }
+    }
+
+    #[tokio::test]
+    async fn test_window_limit_blocks_request_above_limit() {
+        struct FixedCountAnalyticsService {
```
think-in-universe: There're multiple duplicate or similar mocked analytics service in the test, e.g. `FixedCountAnalyticsService`. Can we refactor the code to reduce duplicate code and make the tests easier to maintain?

think-in-universe (approved):

PR #115, opened 2026-01-16
feat: system configs for rate limit

hanakannzashi:


crates/services/src/analytics/ports.rs
```diff
@@ -232,31 +232,4 @@ pub enum AnalyticsError {
...
-    pub fn week() -> Self {
-        Self { seconds: 604800 }
-    }
-
-    /// Month window (2592000 seconds = 30 days)
-    pub fn month() -> Self {
-        Self { seconds: 2592000 }
-    }
-}
+// TimeWindow has been removed, use chrono::Duration directly
```
think-in-universe: Should we remove the comment?

think-in-universe:
@claude review

crates/api/tests/system_configs_tests.rs
```diff
@@ -225,3 +225,405 @@ async fn test_system_configs_write_requires_admin() {
...
+        .get("/v1/configs")
+        .add_header(
+            http::HeaderName::from_static("authorization"),
+            http::HeaderValue::from_str(&format!("Bearer {admin_token}")).unwrap(),
+        )
+        .await;
+
+    assert_eq!(response.status_code(), 200);
+    let body: serde_json::Value = response.json();
+    let rate_limit = body.get("rate_limit").expect("Should have rate_limit");
```
think-in-universe: Can we make the `rate_limit` config only available to admins? It's not necessary to make the rate limit configs accessible to the frontend? It's fine to return all configs with `/v1/admin/configs` but only return a subset of configs in `/v1/configs`.

PR #136, opened 2026-01-29
feat: token usage and cost rate limit

hanakannzashi:


crates/api/src/middleware/rate_limit.rs
```diff
@@ -425,6 +537,30 @@ impl IntoResponse for RateLimitError {
...
+                Some(retry_after_ms),
+            ),
+            RateLimitError::CostLimitExceeded {
+                window_seconds,
+                limit,
+                retry_after_ms,
+            } => (
+                StatusCode::TOO_MANY_REQUESTS,
+                format!(
+                    "Cost limit of {} nano-USD reached for {} second(s) window. Please retry later.",
```
think-in-universe: Can we return USD instead of nano-USD if this error will send to end users?

crates/api/tests/response_author_tests.rs
```diff
@@ -1,490 +0,0 @@
-mod common;
```
think-in-universe: Why remove the response-author mapping tests? I think the features still work, but the implementation is done on cloud-api side. I intentionally keep the tests to make the message authors are still available after #145

crates/database/src/repositories/analytics_repository.rs
```diff
@@ -406,4 +406,69 @@ impl AnalyticsRepository for PostgresAnalyticsRepository {
             })
             .collect())
     }
+
+    async fn record_user_usage(
```
think-in-universe: Should we keep the functions in a separate `usage_respository`?

crates/api/src/routes/api.rs
```diff
@@ -3344,17 +3709,182 @@ async fn proxy_image_edits(
...
+                    .into_response());
+            }
+        };
+        // Use model from request; edits always produce 1 image.
+        let state_clone = state.clone();
+        let user_id = user.user_id;
+        let model = request_model.clone();
+        let mk = services::user_usage::METRIC_KEY_IMAGE_EDIT;
+        tokio::spawn(async move {
+            record_image_usage(&state_clone, user_id, mk, 1, Some(model.as_str())).await;
```
think-in-universe: Why image edits always return 1 image?

hanakannzashi: Because it doesn't support `n` param

crates/api/src/routes/api.rs
```diff
@@ -2425,14 +2431,62 @@ async fn proxy_responses(
...
+    } else if is_streaming_response(&proxy_response.headers) {
+        let usage_stream = UsageTrackingStreamResponseCompleted::new(
+            proxy_response.body,
+            state.user_usage_service.clone(),
+            state.model_pricing_cache.clone(),
+            user.user_id,
+        );
+        Body::from_stream(usage_stream)
+    } else {
+        let bytes = match collect_stream_to_bytes(proxy_response.body).await {
```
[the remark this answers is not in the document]
hanakannzashi: Because `forward_request` always return stream (even if not SSE)

think-in-universe (approved):

PR #144, opened 2026-02-05
docs: add api docs

think-in-universe:
Add missing API docs for endpoints. 

This will be helpful because we're offering access to other projects, especially for non-dev users.

### DIFF
--- a/crates/api/src/routes/api.rs
+++ b/crates/api/src/routes/api.rs
@@ -11,7 +11,7 @@ use axum::{
 use bytes::Bytes;
 use chrono::{Duration, Utc};
 use flate2::read::GzDecoder;
-use http::{header::CONTENT_LENGTH, HeaderName, HeaderValue};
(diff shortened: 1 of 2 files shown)


crates/api/src/routes/api.rs
```diff
@@ -1558,6 +1880,21 @@ async fn clone_conversation(
...
 
 /// Upload a file - forwards to OpenAI and tracks in DB
+#[utoipa::path(
+    post,
+    path = "/v1/files",
+    tag = "Files",
+    request_body(content = Vec<u8>, content_type = "multipart/form-data"),
+    responses(
+        (status = 200, description = "File uploaded successfully", body = crate::models::FileGetResponse),
+        (status = 400, description = "Bad request", body = ErrorResponse),
```
danielwpz: same for the error message, shall we use const instead of repeating the same string?

danielwpz (approved):

PR #147, opened 2026-02-09
chore: build docker image with git commit timestamp

think-in-universe:
Fixes https://github.com/nearai/private-chat/issues/339


think-in-universe:
The test failure of image edit is not relevant to this PR.

PR #164, opened 2026-02-19
Deploy events streaming

nickpismenkov:


henrypark133:
PR #164: Deploy Events Streaming — Review Findings

  #: 1
  Severity: High
  Category: Correctness / Race
  File:Line: service.rs:592-628
  Finding: "created" event without instance field sets created_event_processed = true but never saves instance or cleans up API key. If the Agent API sends {"stage": "created"} without an instance object, the flag is set at line 628, no DB write

    happens, no error is sent, and the API key is permanently orphaned. All subsequent "created" events are ignored as "duplicates".
  Suggested Fix: Move created_event_processed = true inside the if let Some(instance_data) block, and add an else branch that sends an error and revokes the API key.
  ────────────────────────────────────────
  #: 2
  Severity: High
  Category: Race Condition
  File:Line: agents.rs:162-191 vs service.rs:585
  Finding: Instance limit check is TOCTOU-vulnerable, amplified by the streaming path. The limit check (count query at line 162) happens in the route handler, but the DB insert happens asynchronously in a spawned task. Two concurrent SSE
  requests
     can both pass the limit check, then both create instances, exceeding the plan limit. The streaming path widens the race window compared to the synchronous path.
  Suggested Fix: Move the instance-count-and-insert into a serialized transaction, or use a DB-level constraint (e.g., a trigger or INSERT ... WHERE (SELECT count(*) ...) < max). At minimum, re-check the count inside save_instance_from_event
    before inserting.
  ────────────────────────────────────────
  #: 3
  Severity: High
  Category: Resource Leak
  File:Line: service.rs:644-651
  Finding: Orphaned API key + phantom Agent API instance on client disconnect. If the client disconnects before the "created" event, the outer task breaks at line 646, but the inner streaming task (spawned at line 121) keeps running. The Agent
    API may still create the instance upstream. The API key is never revoked, and no DB record is created. The comment at line 649 acknowledges this but defers a fix.
  Suggested Fix: When the outer task breaks due to channel closure before created_event_processed, revoke the API key. Additionally, drop the inner rx to signal the inner task to stop. Consider an AbortHandle on the inner task.
  ────────────────────────────────────────
  #: 4
  Severity: Medium
  Category: DoS / Timeout
  File:Line: service.rs:121-197
  Finding: No idle timeout on the inner streaming task. The 180s reqwest timeout covers the initial connection, but once the stream is established, a slow-drip Agent API can keep the task alive indefinitely by sending 1 byte per minute. This
    leaks a tokio task and holds a channel open.
  Suggested Fix: Add a per-chunk idle timeout using tokio::time::timeout around each stream.next().await, e.g., 60 seconds.
  ────────────────────────────────────────
  #: 5
  Severity: Medium
  Category: Correctness
  File:Line: service.rs:132-146
  Finding: Buffer overflow before size check. buffer.push_str(text) appends the full chunk before the size check at line 135. A single 200KB chunk passes the append, then triggers the error. The buffer momentarily holds 200KB+ of data.
  Suggested Fix: Check buffer.len() + text.len() > MAX_BUFFER_SIZE before calling push_str.
  ────────────────────────────────────────
  #: 6
  Severity: Medium
  Category: Correctness
  File:Line: service.rs:317-384 vs service.rs:535
  Finding: save_instance_from_event bypasses ownership verification on key binding. The non-streaming path calls self.bind_api_key_to_instance(api_key.id, instance.id, user_id) (line 535) which verifies the user owns both the key and instance.
    The streaming path calls repository.bind_api_key_to_instance(*api_key_id, instance.id) directly (line 373), skipping ownership checks. This is probably safe today (both were just created in the same flow), but is fragile if the function is
    reused.
  Suggested Fix: Pass user_id to save_instance_from_event and add an assertion that the created instance's user_id matches, or add a comment explaining why the bypass is safe.
  ────────────────────────────────────────
  #: 7
  Severity: Medium
  Category: API Contract
  File:Line: agents.rs:50-68
  Finding: SSE stream has no terminal event. Clients can't distinguish "stream complete" from "connection dropped by proxy". Standard SSE practice is to send a final event like data: [DONE]\n\n or event: done\n\n.
  Suggested Fix: After the inner ReceiverStream is exhausted, append a sentinel event (e.g., data: {"stage": "done"}\n\n). This likely requires wrapping the stream with .chain(once(...)).
  ────────────────────────────────────────
  #: 8
  Severity: Medium
  Category: Type Safety
  File:Line: service.rs:88, ports.rs:248
  Finding: Trait method returns Receiver<Result<serde_json::Value>> instead of typed events. Both the trait and implementation use untyped serde_json::Value. The LifecycleEvent struct exists in models but is never used. If the Agent API sends
    unexpected shapes, they're forwarded to the client unvalidated.
  Suggested Fix: Either validate/deserialize into LifecycleEvent before forwarding, or at minimum document that the stream is a raw proxy.
  ────────────────────────────────────────
  #: 9
  Severity: Low
  Category: Correctness
  File:Line: agents.rs:194-198
  Finding: Only checks the first Accept header value. headers.get("accept") returns only the first header. Clients can send multiple Accept headers or a single comma-delimited one like text/event-stream, application/json. The current code
  handles
    the comma-delimited case (via .contains()) but misses multiple headers.
  Suggested Fix: Use headers.get_all("accept") and check all values.
  ────────────────────────────────────────
  #: 10
  Severity: Low
  Category: Robustness
  File:Line: service.rs:131
  Finding: Invalid UTF-8 chunks are silently dropped. std::str::from_utf8(&chunk) returns Err for invalid UTF-8, and the code silently skips the entire chunk. A multi-byte UTF-8 character split across chunk boundaries would cause data loss.
  Suggested Fix: Use String::from_utf8_lossy or accumulate raw bytes and decode at line boundaries.
  ────────────────────────────────────────
  #: 11
  Severity: Nit
  Category: Style
  File:Line: agents.rs:231-243
  Finding: Non-streaming JSON path manually builds Response. The original (StatusCode::CREATED, Json(instance.into())) was simpler and more idiomatic for axum. The manual path loses automatic content-type negotiation.
  Suggested Fix: Consider using (StatusCode::CREATED, Json(InstanceResponse::from(instance))).into_response() and returning Response from that.
  ────────────────────────────────────────
  #: 12
  Severity: Nit
  Category: Improvement
  File:Line: usage_parsing.rs:200,293
  Finding: buffer.drain(..=last_newline) is a good improvement over the previous this.buffer[last_newline + 1..].to_string() — avoids an allocation. Correct change.
  Suggested Fix: N/A — looks good.

PR #165, opened 2026-02-19
feat: route model "auto" to zai-org/GLM-5-FP8

henrypark133:
## Summary
- When `model: "auto"` is sent to `/v1/chat/completions`, substitute it with `zai-org/GLM-5-FP8` and inject default parameters (`temperature=1.0`, `top_p=0.95`, `max_tokens=4096`) if not already provided
- Added integration tests using wiremock to verify the substituted model and default parameters in the forwarded upstream request

## Test plan
- [x] `test_chat_completions_auto_model_not_blocked` — verifies `model: "auto"` is not rejected by visibility check
- [x] `test_chat_completions_auto_model_forwards_with_substituted_model` — verifies forwarded request has `zai-org/GLM-5-FP8` model and correct default params via wiremock
- [x] Manual curl test against staging API confirmed model and defaults work

### DIFF
--- a/build-config.toml
+++ b/build-config.toml
@@ -8,7 +8,7 @@
 #   - "main" (branch)
 #   - "v1.2.3" (tag)
 #   - "abc123def456" (commit hash)
-private_chat_frontend_version = "5353664990bfa21712c15884cd06e523ac047eba" # v0.1.13
+private_chat_frontend_version = "3b487faa7b9a5717911254aa670ddc3b562475ed" # v0.2.0-dev
 
 # PostHog key and host for the private-chat frontend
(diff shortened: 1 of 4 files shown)


henrypark133:
All review comments addressed in 37d5b64:

1. **Hardcoded magic values** → Extracted into `pub const AUTO_ROUTE_*` constants
2. **`is_none()` doesn't handle JSON `null`** → Changed to `is_none_or(|v| v.is_null())`
3. **Weak first test** → Removed; the wiremock-based test already covers visibility + full round-trip
4. **Redundant inner `if let`** → Simplified
5. **No info log** → Added `tracing::info!("Auto-routing model: user_id={}", ...)`

Also added a new test (`test_chat_completions_auto_model_preserves_client_params`) that verifies client-provided `temperature`/`max_tokens` are preserved while missing params get defaults.

PR #182, opened 2026-02-20
feat: purchase tokens

### LINKED ISSUE #180: Feature: purchase tokens
Allow chat-api users to purchase tokens with Stripe, so users can add more token balances to their agent when needed.

The total token balances of the user will be the purchased tokens + the max monthly tokens for each subscription plan. When the consumed tokens in this month exceeds the max monthly tokens for each subscription plan, the purchased token balance will be consumed.

think-in-universe:
Resolves #180


think-in-universe:
Close in favor of #178

PR #186, opened 2026-02-23
feat: add instance upgrade endpoint

PierreLeGuen:
## Summary

- Adds `POST /v1/agents/instances/{id}/upgrade` endpoint that upgrades an instance to the latest image for its service type
- The endpoint fetches current images from the owning compose-api's `/version` endpoint, maps `service_type` to the correct image key (worker/ironclaw), and restarts with that digest
- Includes OpenAPI docs and follows existing restart endpoint patterns

## Test plan

- [x] Deploy chat-api and verify `POST /v1/agents/instances/{id}/upgrade` returns 200
- [x] Verify the instance restarts with the latest image digest from compose-api
- [x] Verify ownership check — upgrading another user's instance returns 403

### DIFF
--- a/crates/api/tests/agent_tests.rs
+++ b/crates/api/tests/agent_tests.rs
@@ -262,13 +262,11 @@ async fn test_create_instance_rejects_unsubscribed_user() {
     let user_email = "no_sub_user@example.com";
     let user_token = mock_login(&server, user_email).await;
 
-    // Clean up any leftover subscription plans from other tests
-    common::clear_subscription_plans(&server).await;
-
     // Ensure user has NO subscription
     common::cleanup_user_subscriptions(&db, user_email).await;
 
     // Set subscription plans (for subscribed users); unsubscribed users get 0 instances
+    // Note: set_subscription_plans overwrites any existing plans, so no need to clear first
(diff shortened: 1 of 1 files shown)


henrypark133:
### Code review

Found 2 issues:

1. `instance.name` is logged at `warn` level (production-enabled) in `check_upgrade_available`, violating CLAUDE.md which says to never log \"Metadata that reveals customer information\" and to log \"IDs only\" at production log levels.

https://github.com/nearai/chat-api/blob/3cd823815a7476b4b0378d41662da05a52b72554/crates/services/src/agent/service.rs#L1458-L1463

2. `check_upgrade_available` is missing from the OpenAPI spec registration in `openapi.rs`. `upgrade_instance` was added but its paired endpoint was not, so it won't appear in generated API docs.

https://github.com/nearai/chat-api/blob/3cd823815a7476b4b0378d41662da05a52b72554/crates/api/src/openapi.rs#L101-L107

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

henrypark133 (approved):

PR #191, opened 2026-02-25
fix: agent instance status and sanitize dashboard url

think-in-universe:
Fixes

1. The agent `status` in db is not updated when user operates;
2. The `dashboard_url` should be sanitized in the admin `/v1/agents/instances` endpoint to avoid leaking access info;


crates/services/src/agent/service.rs
```diff
@@ -1250,6 +1250,11 @@ impl AgentService for AgentServiceImpl {
             ));
         }
 
+        // Update DB status to active (end state after restart)
+        self.repository
+            .update_instance_status(instance_id, "active")
```
danielwpz: should we check the instance status is not 'active' at the start of this restart_instance service method?

PR #197, opened 2026-02-27
fix: remove Agent API enrichment from list instances endpoint

Evrard-Nil:
## Summary
- Removed `get_instance_enrichments()` call from `GET /v1/agents/instances` which was making HTTP calls to external Agent API managers on every request, causing ~15s response times
- The endpoint now returns DB data directly via `InstanceResponse::from()`, responding in milliseconds
- Status is already kept in sync by the background `sync_all_instance_statuses` job
- The single-instance endpoint (`GET /v1/agents/instances/{id}`) still does enrichment for `ssh_command`

## Test plan
- [ ] Verify `GET /v1/agents/instances` responds in <1s
- [ ] Verify instance status is correct in list (from DB, synced by background job)
- [ ] Verify `GET /v1/agents/instances/{id}` still returns `ssh_command` and live status
- [ ] Companion frontend PR: nearai/private-assistant (DeployedView fetches individual instance for ssh_command)


think-in-universe:
The agent status is not yet synchronized for production. There was a bug of agent status that is fixed in #191, but we still need to sync the statuses of all agents after deploy to production.

I think we also need to save the SSH command in chat-api (needs DB migration), since it shouldn't change after the instance is created. The current fix in the PR works well as a workaround, but we need a complete fix soon.

PR #202, opened 2026-03-02
fix: skip near balance check for paid users

hanakannzashi:


crates/services/src/system_configs/ports.rs
```diff
@@ -78,6 +78,9 @@ pub struct RateLimitConfig {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct PaymentProviderConfig {
     pub price_id: String,
+    /// Price in cents (e.g. 999 for $9.99, 0 for free)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub price: Option<i64>,
```
think-in-universe: We should add the `price` field to `SubscriptionPlanConfig`. It should be consistent for the plans. No specific config for each provider.

hanakannzashi: I think price shouldn’t be in SubscriptionPlanConfig.

The same plan can have different prices across providers (e.g., Apple is usually more expensive), so price is provider-specific, not plan

PR #210, opened 2026-03-06
feat(analytics): enhance user, agent and usage data queries

think-in-universe:
## Summary
Enhances admin BI analytics: user subscription handling, user summary and deployment list behaviour, usage aggregations, and production safeguards (timeouts, stable responses, robust column access).
---
### User subscription & BI users
- **Active subscription only:** `/v1/admin/bi/users` now attributes a user’s subscription only when it is **active** or **trialing**. Subscription lookup filters `status IN ('active', 'trialing')` and selects by `current_period_end DESC` so the “current” plan is unambiguous when a user has multiple subscriptions.
### User summary (new)
- **New endpoint:** `GET /v1/admin/bi/users/summary` returns:
  - `by_subscription_plan`: user count per resolved plan name (from system config; unknown price IDs → `"Unknown"`).
  - `by_agent_count`: user count per deployed-agent-count bucket.
- Plan names are aggregated after resolving price ID → plan (multiple price IDs for the same plan, or multiple unknowns, are merged into one bucket).
- **Stable ordering:** `by_subscription_plan` is sorted by `user_count` desc, then `plan` asc to avoid non-deterministic API order and UI flicker.
- **Single query:** One CTE pass computes both aggregations (no double full scan of users/agent_instances/subscriptions).
- **Statement timeout:** Summary query runs inside a transaction with `SET LOCAL statement_timeout = '30000'` (same as other heavy BI queries).
### Deployments (agents) list
- **Response fields:** `DeploymentRecord` (and `GET /v1/admin/bi/deployments`) now includes:
  - `name` (agent instance name),
  - `total_spent_nano` (sum of `cost_nano_usd` for the instance),
  - `total_tokens` (sum of token usage only: `metric_key = 'llm.tokens'`), so non-token metrics (e.g. images) are not counted.
- **Search & sort:** Query params `q` (search), `sort_by`, and `sort_order` are supported. Search is case-insensitive over agent name, user email, and user name; `%` and `_` in `q` are escaped so they are treated literally (no SQL wildcard behaviour). Sort options include `total_spent_nano` and `total_tokens`.
- **Validation:** `DeploymentsSortBy` uses `FromStr` for validation/parsing. The previous “too many wildcard characters” check for `%` in `q` was removed so valid searches containing literal `%` are not rejected.
- **Performance:** Usage totals per deployment are computed with a **LATERAL** join (per-row aggregate on `user_usage_event` for that `instance_id`) instead of a single global aggregate, to avoid a full table scan.
- **Robust mapping:** Deployment and usage rows use **named column access** in the repository (no positional indices) so adding or reordering columns does not break the mapper.
### Usage aggregation
- **Active counts:** When `group_by` is `day`, `model`, or `user`, the usage response includes `active_agents_count` and `active_users_count` (distinct agents/users with usage &gt; 0 in that group). These are read by **named column** in the repo.
- Existing usage behaviour (tokens, cost, request counts) is unchanged.
### Production safety
- **Statement timeouts:** `get_user_summary` and `list_users_with_stats` run inside a transaction with `SET LOCAL statement_timeout = '30000'`, consistent with `get_usage_aggregation` and `get_status_history`, so long-running BI queries do not hold connections indefinitely.
### Tests
- **User summary:** `test_bi_users_summary` seeds agent instances and asserts `GET /v1/admin/bi/users/summary` returns `by_subscription_plan` and `by_agent_count` with the expected shape and at least one user in the agent-count distribution.
- **Usage:** `test_bi_usage_group_by_day_with_data` asserts `active_agents_count` and `active_users_count` are present and &gt; 0 when usage data exists.
- **Deployments:** `test_bi_list_deployments_with_data` asserts each deployment has `name`, `total_spent_nano`, and `total_tokens`. New `test_bi_list_deployments_search_and_sort` checks that `q` filters by name and that `sort_by=name&sort_order=asc` returns deployments in ascending name order.
---
## API
| Method | Path | Description |
|--------|------|-------------|
| GET | `/v1/admin/bi/users` | List users with stats; subscription is active/trialing only. |
| GET | `/v1/admin/bi/users/summary` | **New.** User counts by subscription plan and by agent count. |
| GET | `/v1/admin/bi/deployments` | List deployments with `q`, `sort_by`, `sort_order`; response includes `name`, `total_spent_nano`, `total_tokens`. |
| GET | `/v1/admin/bi/usage` | With `group_by=day` (or `model`/`user`), response includes `active_agents_count` and `active_users_count`. |
---
## Checklist
- [x] Active/trialing-only subscription for BI users
- [x] User summary endpoint with stable ordering and single-query CTE
- [x] Deployment list: name, total_spent_nano, total_tokens, search, sort
- [x] Usage: active_agents_count / active_users_count and named column access
- [x] Statement timeouts on get_user_summary and list_users_with_stats
- [x] Removed misleading “wildcard” validation for deployment search
- [x] Tests for user summary, usage active counts, and deployment fields/search/sort

### DIFF
--- a/crates/database/src/repositories/bi_metrics_repository.rs
+++ b/crates/database/src/repositories/bi_metrics_repository.rs
@@ -460,12 +460,12 @@ impl BiMetricsRepository for PostgresBiMetricsRepository {
             .into_iter()
             .map(|r| {
                 let active_agents_count = if include_active_agents {
-                    Some(r.get::<_, i64>(12))
+                    Some(r.get::<_, i64>("active_agents_count"))
                 } else {
                     None
                 };
                 let active_users_count = if include_active_users {
-                    Some(r.get::<_, i64>(13))
+                    Some(r.get::<_, i64>("active_users_count"))
                 } else {
                     None
                 };
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -1981,17 +1981,10 @@ impl BiDeploymentQuery {
     fn validate_deployments(&self) -> Result<(), ApiError> {
         if let Some(ref q) = self.q {
             let q_trimmed = q.trim();
-            if !q_trimmed.is_empty() {
-                if q_trimmed.len() > 200 {
-                    return Err(ApiError::bad_request(
-                        "search query exceeds maximum length of 200",
-                    ));
-                }
-                if q_trimmed.matches('%').count() > 5 {
-                    return Err(ApiError::bad_request(
-                        "search query contains too many wildcard characters",
-                    ));
-                }
+            if !q_trimmed.is_empty() && q_trimmed.len() > 200 {
+                return Err(ApiError::bad_request(
+                    "search query exceeds maximum length of 200",
+                ));
             }
         }
         if self.sort_by.parse::<DeploymentsSortBy>().is_err() {
--- a/crates/api/tests/bi_metrics_tests.rs
+++ b/crates/api/tests/bi_metrics_tests.rs
@@ -227,7 +227,7 @@ async fn test_bi_list_deployments_with_data() {
         "Deployments array should not be empty"
     );
 
-    // Verify response fields are present on each deployment (including user info for admin UI)
+    // Verify response fields are present on each deployment (including user info and usage-derived fields)
     for d in deployments {
         assert!(d.get("id").is_some());
         assert!(d.get("user_id").is_some());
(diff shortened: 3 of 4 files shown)


think-in-universe:
@claude review

crates/database/src/repositories/bi_metrics_repository.rs
```diff
@@ -343,20 +450,41 @@ impl BiMetricsRepository for PostgresBiMetricsRepository {
...
-                input_tokens: r.get(4),
-                output_tokens: r.get(5),
-                total_tokens: r.get(6),
-                input_cost_nano: r.get(7),
-                output_cost_nano: r.get(8),
-                total_cost_nano: r.get(9),
-                request_count: r.get(10),
+            .map(|r| {
+                let active_agents_count = if include_active_agents {
+                    Some(r.get::<_, i64>(12))
```
hanakannzashi: Using string keys is less error-prone than using numeric keys.

PR #229, opened 2026-03-25
feat: implement cronjob service and scheduler

danielwpz:
Deployment notes for the new scheduled cleanup task system:

  ### Runtime model
  We should deploy this as **one image with two containers**:
  - `chat-api` runs `./api`
  - `chat-api-worker` runs `./task_worker`

  The API process is responsible for **creating/ensuring** the daily cleanup schedule at startup.
  The worker process is responsible for **consuming** scheduled tasks from SQS and executing the cleanup.

  ### Docker/image changes
  The image now includes both binaries:
  - `api`
  - `task_worker`

  The default container command remains `./api`, so the worker container should override the command to:

  ```yaml
  command: ["./task_worker"]
  ```
  First-time infrastructure/config required

  Before the first deployment, we need:
  - an SQS queue for task delivery
  - an AWS Scheduler role that can send messages to that SQS queue
  - AWS credentials/permissions for the API container to create/update schedules
  - AWS credentials/permissions for the worker container to receive/delete SQS messages

  API container config

  The API container needs task scheduler config:

  TASKS_ENABLED=true
  TASKS_AWS_REGION=<aws-region>
  TASKS_SQS_QUEUE_ARN=<sqs-queue-arn>
  TASKS_SCHEDULER_ROLE_ARN=<scheduler-role-arn>
  TASKS_SCHEDULER_GROUP=chat-api

  Optional cleanup config:

  TASKS_CLEANUP_CANCELED_INSTANCES_DAILY_CRON=cron(0 0 * * ? *)
  TASKS_CLEANUP_CANCELED_INSTANCES_GRACE_DAYS=15

  Defaults:
  - cron: cron(0 0 * * ? *) (UTC midnight every day)
  - grace days: 15

  Worker container config

  The worker container needs task worker config:

  TASKS_ENABLED=true
  TASKS_AWS_REGION=<aws-region>
  TASKS_SQS_QUEUE_URL=<sqs-queue-url>
  TASKS_WORKER_WAIT_SECONDS=20
  TASKS_WORKER_VISIBILITY_TIMEOUT=60
  TASKS_WORKER_MAX_MESSAGES=10
  TASKS_WORKER_MAX_CONCURRENCY=10

  It also still needs the existing runtime config required to delete instances:
  - database config
  - agent manager config
  - the same shared init/config volume setup used by the API container

  First deployment checklist

  1. Deploy the updated image containing both api and task_worker
  2. Add a dedicated chat-api-worker container using the same image
  3. Mount the same init-generated config volume into the worker container
  4. Configure task scheduler env on the API container
  5. Configure task worker env on the worker container
  6. Start both containers

  Important note

  Right now the API uses create_task_if_absent(...) when ensuring the daily cleanup schedule.

  That means the new cron/grace-day config is applied correctly on first creation, but changing these env vars later will not automatically update an existing schedule.
  If we want post-deploy config changes to propagate automatically, we should switch that path to an upsert/update flow later.


danielwpz:
@claude 
3： This is a real pagination gap: the cleanup currently only reads the first 1,000 instances for a user. Given our current configured instance limits, it’s unlikely to affect normal production users, so I’m not treating it as a blocker for this fix set. I’d prefer to handle the inner pagination separately in a follow-up change.

other issues are fixed, please review again

crates/api/src/tasks/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+    if let Some(receipt_handle) = message.receipt_handle() {
+        client
+            .delete_message()
+            .queue_url(queue_url)
+            .receipt_handle(receipt_handle)
+            .send()
+            .await
+            .context("failed to delete SQS message")?;
+    }
+
```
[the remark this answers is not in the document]
danielwpz: I’m not changing this one in this PR. If receipt_handle is missing, we can’t delete the message either way, so returning an error instead of Ok(()) doesn’t actually change the redelivery behavior. In normal SQS ReceiveMessage responses a receipt handle is expected, so I’m treating this as an edge-case defensive improvement rather than a priority fix.

crates/api/src/tasks/mod.rs
```diff
@@ -0,0 +1,553 @@
...
+        let semaphore = Arc::new(Semaphore::new(self.max_concurrency.max(1)));
+
+        loop {
+            let response = match self
+                .client
+                .receive_message()
+                .queue_url(&self.queue_url)
+                .max_number_of_messages(self.max_messages.clamp(1, 10))
+                .wait_time_seconds(self.wait_seconds.clamp(1, 20))
+                .visibility_timeout(self.visibility_timeout.max(1))
```
hanakannzashi: We set the SQS visibility timeout once when receiving the message, but we never extend it while `dispatch_task` is running. This cleanup job can easily exceed 60 seconds, which means the same message may become visible and be processed by another worker before the first one deletes it. Can we heartbeat `ChangeMessageVisibility` (or otherwise guarantee the handler finishes within the visibility window)?

danielwpz: Visibility heartbeat is a reasonable robustness improvement, but implementing it correctly is a non-trivial change to the worker execution model rather than a small fix. It adds background renewal, cancellation, and message lifecycle coordination, which increases complexity and introduces new failure modes. For the current cleanup task, duplicate execution is low risk because the delete path is effectively replay-safe, so I’d rather handle heartbeat support as a focused follow-up instead of folding it into this PR.

Issue #226, opened 2026-03-23
Issue: Subscription Cancellation Incorrectly Carries Over Credits to Free Plan


sunglow666:
### Description
When a Starter subscription is canceled and the user transitions to the Free Plan, the system incorrectly carries over usage credits from the previous subscription period, resulting in unexpected consumption of the user's purchased balance.

### Affected User
user_id = 942cdc8b-e787-48d2-916b-fc3a67e8b8f0 (production)

### Timeline

1. [2026-02-22T08:22:16Z] User subscribes to Starter plan

  - plan_credits: $5.00 (5,000,000,000 nano)
  - Usage: $5.01477265, overspent by $0.01477265

```
{
  "balance": 0,
  "period_spent_credits": 5014772650,
  "plan_credits": 5000000000
}
```

2. [2026-03-01T13:46:39Z] User requests cancellation
3. [2026-03-20] User adds $1.00 purchased credits

  - Overspent $0.01477265 deducted from purchased balance
  - Remaining purchased balance: $0.98522735
```
{
  "balance": 985227350,
  "total_purchased_nano_usd": 1000000000,
  "spent_purchased_nano_usd": 14772650
}
```

4. Additional usage consumes more purchased credits
```
{
  "balance": 968523950,
  "spent_purchased_nano_usd": 31476050
}
```

5. [2026-03-22T08:22:16Z] Subscription cancellation takes effect – User transitions to Free Plan

- **Unexpected behavior:**

  - Free Plan shows plan_credits: $2.00 (2,000,000,000 nano)
  - period_spent_credits: $4.61400555 (should be $0 for Free Plan ?)
  - **All purchased balance ($0.96852395) was unexpectedly consumed**
```
{
    "balance": 0,
    "total_purchased_nano_usd": 1000000000,
    "spent_purchased_nano_usd": 1000000000,
    "period_spent_credits": 4614005550,
    "plan_credits": 2000000000
}
```

### Expected Behavior

- **Upon cancellation and transition to Free Plan:**

  - period_spent_credits should reset to 0 ?
  - Previously accumulated usage from the Starter period should not affect the Free Plan
  - Purchased balance should not be consumed to cover usage from a prior subscription period

PR #230, opened 2026-03-25
fix: free plan period after subscription plan cancelled

think-in-universe:
Fix https://github.com/nearai/chat-api/issues/226


think-in-universe:
@claude please review again

PR #246, opened 2026-04-03
fix: convert service types for non-tee hosting

think-in-universe:


crates/services/src/agent/service.rs
```diff
@@ -1134,7 +1144,7 @@ impl AgentServiceImpl {
             "nearai_api_key": nearai_api_key,
             "nearai_api_url": nearai_api_url,
             "ssh_pubkey": params.ssh_pubkey,
-            "service_type": service_type_for_api,
+            "service_type": agent_api_service_type,
```
danielwpz: as the variable name suggests, here `service_type_for_api` shall be used right? maybe we should swap the name of these two vars

hanakannzashi (approved):

danielwpz (approved):

PR #248, opened 2026-04-08
fix: check available upgrade with docker image digest and version tags

nickpismenkov:


think-in-universe:
@claude please help review

crates/services/src/agent/service.rs
```diff
@@ -4129,47 +4146,116 @@ impl AgentServiceImpl {
...
 
-        // Determine upgrade availability by comparing semantic versions
-        let current_version = extract_version_from_image(current_image.as_ref().unwrap());
-        if let Some(ref v) = current_version {
+        let (current_image_ref, current_digest) = current_image.unwrap();
+
+        // Determine upgrade availability based on image tag type
+        let current_version = extract_version_from_image(&current_image_ref);
+
+        if let Some(curr_v) = current_version {
```
think-in-universe, line 4408: From user experience's perspective, I think we should always upgrade to the latest versioned image tag if applicable, which is more readable and easier to understand by users.

- If the current image is `latest` or `staging`, upgrade to the latest version (e.g. `v0.24.0`) if image digests are different and the target version has a newer image (find the versioned tag of `latest` or `staging` and then compare the versions, or compare creation date if available);
- If the current image is versioned tag (e.g. `v0.23.0`), upgrade to the latest version (e.g. `v0.24.0`) if image digests are different and the target version is newer.

**Edit**: after discussion, I think we can keep the current logic unchanged:

1. `staging` only upgrades to the latest `staging` tag; `latest` tag in practice won't be used for deployment because we only deploy version tag in production;
2. versioned tag only upgrades to versioned tag;

think-in-universe (changes requested):
With the current versioning strategy, I think we should upgrade to a versioned tag even if the current image tag is non-versioned. 

**Edit**: see below comment for more details. 

- mutable tag upgrade to mutable tag
- immutable version tag upgrade to immutable version tag

think-in-universe (commented):
Two questions:

1. We can send image digest to crabshack for instance creation or upgrade request, but let's make sure the image version of agent instance is still available somewhere if image digest is used for agent creation or upgrade, so we can show the current version number in frontend. Is this possible?
2. For semver comparison, have we tested more cases like `1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0`? For production, we should only allow upgrading to stable versions, but we may be more flexible in staging.

think-in-universe:

think-in-universe (changes requested):

crates/services/src/agent/service.rs
```diff
@@ -4129,47 +4272,118 @@ impl AgentServiceImpl {
...
+                matching_entry.and_then(|e| e.image_digest.as_ref()),
+                has_upgrade
+            );
 
-        Ok(UpgradeAvailability {
-            has_upgrade,
-            current_image,
-            latest_image,
-        })
+            Ok(UpgradeAvailability {
```
think-in-universe, line 4515: Can we include optional `current_digest` and `latest_digest` in the response as well? So the frontend can show the different image digests for staging when needed.

crates/services/src/agent/service.rs
```diff
@@ -4263,12 +4474,88 @@ impl AgentServiceImpl {
...
+
+            tracing::debug!(
+                "Non-TEE upgrade: Non-versioned tag found in allowlist: ref={}, current_digest={:?}, allowlist_digest={:?}",
+                current_image,
+                current_instance.image_digest,
+                allowlist_entry.and_then(|e| e.image_digest.as_ref())
+            );
+
+            // Use the same image ref - manager will pull the latest digest of this tag
+            current_image.clone()
```
think-in-universe: Ideally we should send the target image digest to crabshack (crabshack already supports this) so there's no ambiguity.

think-in-universe (changes requested):

crates/services/src/agent/service.rs
nickpismenkov: added

think-in-universe:

think-in-universe:
@claude please review again

think-in-universe:
@claude please review again

think-in-universe:
@claude please review again

think-in-universe (approved):

PR #249, opened 2026-04-08
fix: agent usage data for non-LLM service

hanakannzashi:


think-in-universe:

think-in-universe (changes requested):
In agent usage endpoint responses, we should return the actual quantity for the specific usage type, otherwise natively removing the quantity from the result for non-LLM usage type.

We probably need to update the response structure to include missing data field.

```rust
/// Usage log entry response
#[derive(Debug, Serialize, Deserialize, ToSchema)]
pub struct UsageResponse {
    pub id: String,
    pub api_key_name: String,
    pub input_tokens: i64,
    pub output_tokens: i64,
    pub total_tokens: i64,
    pub input_cost: String,  // nano-dollars formatted as string
    pub output_cost: String, // nano-dollars formatted as string
    pub total_cost: String,  // nano-dollars formatted as string
    pub model_id: String,
    pub request_type: String,
    pub created_at: String,
}
```

think-in-universe (approved):

PR #256, opened 2026-04-10
fix: preserve image tag for agent instance upgrade

think-in-universe:
In the current implementation, the image tag is lost. e.g. `docker.io/nearaidev/ironclaw-dind` is kept in the DB after upgrade. `staging` tag is lost.

Preserve the image tag for agent upgrading request.


think-in-universe:
@claude please review again

hanakannzashi (approved):

PR #258, opened 2026-04-13
feat(admin): agent instance status history with more details

think-in-universe:


hanakannzashi (approved):

think-in-universe:

PR #262, opened 2026-04-17
fix: handle sse streaming failure

nickpismenkov:
Summary of changes to crates/services/src/agent/service.rs:                                                                                 
                                                                                                                                              
  Fixed SSE stream error handling to correctly distinguish between:                                                                           
  1. Normal connection closure (buffer empty) → silently treat as successful completion                                                       
  2. Real stream errors (unprocessed data) → propagate the error 
  
  This ensures operations that complete successfully (all SSE events received and processed) aren't marked as failed just because the         
  connection closed afterward.

### DIFF
--- a/crates/services/src/agent/service.rs
+++ b/crates/services/src/agent/service.rs
@@ -1473,17 +1473,18 @@ impl AgentServiceImpl {
                         }
                     }
                     Some(Err(e)) => {
-                        // If stream ended after successfully processing all received data,
-                        // treat as normal stream closure (connection terminated after sending events)
-                        if chunk_count > 0 && buffer.is_empty() {
+                        // If all received data was successfully processed (buffer is empty),
+                        // the stream closed normally after completion
+                        if buffer.is_empty() {
                             break;
                         }
 
+                        // Unprocessed data remains - this is a real error
                         tracing::error!(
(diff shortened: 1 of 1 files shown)


crates/services/src/agent/service.rs
```diff
@@ -1463,7 +1473,19 @@ impl AgentServiceImpl {
                         }
                     }
                     Some(Err(e)) => {
-                        tracing::error!("Error reading Agent API stream: {}", e);
+                        // If all received data was successfully processed (buffer is empty),
+                        // the stream closed normally after completion
+                        if buffer.is_empty() {
+                            break;
```
henrypark133, line 1479: I'm not very familiar with the flow but this seems very hacky. If I'm reading the code and description correctly, an error with empty buffer is essentially being treated as success. We need some final event that confirms that it is actually done ideally. Can you confirm that this is the intended change?

nickpismenkov: yep, that's kinda hacky, but it would fix the prod. The actual problem is that on the crabshack side it doesn't fire the final event properly. But I think it's better to unblock it with this fix for now

henrypark133 (approved):

PR #271, opened 2026-04-28
feat: implement async user account deletion

danielwpz:
## Summary
- Add async DELETE /v1/users/me account deletion workflow
- Block deleting users with active subscriptions or non-stopped instances
- Add account deletion worker task handling and progress tracking
- Combine FK cascade changes and deletion request table into one V34 migration

## Verification
- cargo fmt --check
- CARGO_INCREMENTAL=0 cargo clippy --workspace --all-targets --features test -- -D warnings
- cargo test -p api --features test --test user_account_deletion_tests


crates/api/src/middleware/auth.rs
```diff
@@ -246,6 +280,10 @@ pub async fn auth_middleware(
     let user = authenticate_token_string(token, &state)
         .await
         .map_err(|e| e.into_response())?;
+    let allow_delete_retry = method == Method::DELETE && (path == "/v1/users/me" || path == "/me");
```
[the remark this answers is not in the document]
danielwpz: Addressed in `62e7d7e`.

I normalized the request path before comparison using `trim_end_matches(/)`, so delete-retry allowance now works with trailing slash variants as well.

crates/api/src/middleware/auth.rs
```diff
@@ -199,6 +200,39 @@ async fn authenticate_session_by_token(
...
+    allow_delete_retry: bool,
+) -> Result<(), ApiError> {
+    if allow_delete_retry {
+        return Ok(());
+    }
+
+    let is_deleting = state
+        .user_service
+        .is_account_deletion_requested(user_id)
+        .await
```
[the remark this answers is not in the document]
danielwpz: Good point on the auth-path query cost.

I did not introduce a TTL cache in this patch to keep behavior changes minimal for this release, but I agree with the direction and will handle caching/invalidation as a focused follow-up (separate PR) so we can validate hit-rate and staleness behavior safely.

crates/database/src/repositories/user_repository.rs
```diff
@@ -139,16 +227,329 @@ impl UserRepository for PostgresUserRepository {
...
+    ) -> anyhow::Result<()> {
+        let client = self.pool.get().await?;
         client
-            .execute("DELETE FROM users WHERE id = $1", &[&user_id])
+            .execute(
+                "UPDATE user_account_deletions
+                 SET progress = $2, updated_at = NOW()
+                 WHERE id = $1",
+                &[&deletion_id, &progress],
+            )
```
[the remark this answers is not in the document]
danielwpz: Addressed in `62e7d7e`.

`update_account_deletion_progress` now refreshes `lease_until` on each progress write, and the worker passes `ACCOUNT_DELETION_LEASE_SECONDS` for every update. This prevents lease expiry during long conversation cleanup loops.

crates/api/src/middleware/auth.rs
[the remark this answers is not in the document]
danielwpz: Correction for the snippet in my previous reply: path normalization uses `trim_end_matches("/")`.

danielwpz:
Pushed fix commit `62e7d7e` addressing the main review points:

- Added bounded retry handling for account deletion tasks.
  - New worker guard: `MAX_ACCOUNT_DELETION_ATTEMPTS`.
  - Exceeded attempts now transition to `failed_needs_review` (with `last_error`) instead of retrying forever.
- Added lease refresh on progress updates.
  - `update_account_deletion_progress` now also extends `lease_until` to reduce long-job lease expiry races.
- Hardened auth delete-retry path matching via normalized path (`trim_end_matches("/")`).
- Updated `is_account_deletion_requested` to treat `completed` as not deleting.

I intentionally deferred the per-request deletion-status cache optimization to a follow-up PR to keep this release focused on correctness/safety fixes.

crates/api/src/routes/users.rs
```diff
@@ -41,6 +44,104 @@ pub async fn get_current_user(
...
+        "Requesting current user account deletion: user_id={}",
+        user.user_id
+    );
+
+    app_state
+        .user_service
+        .validate_account_deletion_preconditions(user.user_id)
+        .await
+        .map_err(account_deletion_error_to_api_error)?;
+
```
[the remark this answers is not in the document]
danielwpz: Good catch - removed the redundant validate_account_deletion_preconditions call in the handler. create_account_deletion_request already validates atomically inside its own transaction with FOR UPDATE.

crates/database/src/repositories/user_repository.rs
```diff
@@ -139,16 +227,355 @@ impl UserRepository for PostgresUserRepository {
...
+                "INSERT INTO user_account_deletions (user_id, status)
+                 VALUES ($1, 'pending')
+                 ON CONFLICT (user_id) DO UPDATE
+                    SET updated_at = NOW()
+                 RETURNING id, user_id, status, requested_at, started_at, completed_at,
+                           attempt_count, lease_until, last_error, progress, created_at, updated_at",
+                &[&user_id],
+            )
+            .await
+            .map_err(anyhow::Error::from)?;
```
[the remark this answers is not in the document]
danielwpz: Valid concern. Added a status check after the INSERT ON CONFLICT: if the existing row is failed_needs_review or completed, the handler now returns a 409 Conflict (AlreadyInTerminalState) instead of a misleading 202.

crates/database/src/repositories/user_repository.rs
```diff
@@ -12,6 +18,88 @@ impl PostgresUserRepository {
...
+        if !blocking_instance_rows.is_empty() {
+            let mut total = 0_i64;
+            let mut statuses = Vec::with_capacity(blocking_instance_rows.len());
+            for row in blocking_instance_rows {
+                let status: String = row.get("status");
+                let count: i64 = row.get("count");
```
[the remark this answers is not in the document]
danielwpz: The deletion is asynchronous (SQS worker), so a FOR UPDATE on agent_instances during validation would not meaningfully close the window - the transaction commits immediately and the worker may execute minutes later. The precondition check here is a best-effort guard; the worker already re-checks instance state in delete_user_account. Not adding instance-level locking.

crates/api/src/bin/task_worker.rs
```diff
@@ -180,6 +209,157 @@ impl TaskExecutor for DefaultTaskExecutor {
...
+            request.id,
+            request.user_id,
+            request.attempt_count
+        );
+
+        if request.attempt_count > MAX_ACCOUNT_DELETION_ATTEMPTS {
+            let last_error = format!(
+                "max deletion retry attempts exceeded (attempt_count={}, max={})",
+                request.attempt_count, MAX_ACCOUNT_DELETION_ATTEMPTS
+            );
```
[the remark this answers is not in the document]
danielwpz: The analysis here is not quite right. claim_account_deletion does attempt_count = attempt_count + 1 before the check attempt_count > MAX. With MAX=5, the trace is: 0-1 (run), 1-2 (run), ... 4-5 (run), 5-6 (fail). That is exactly 5 executions, not MAX+1. The > operator is correct here.

think-in-universe:

crates/database/src/repositories/user_repository.rs
```diff
@@ -139,16 +227,367 @@ impl UserRepository for PostgresUserRepository {
...
+            &[&user_id],
+        )
+        .await
+        .map_err(anyhow::Error::from)?;
+
+        tx.execute("DELETE FROM conversations WHERE user_id = $1", &[&user_id])
+            .await
+            .map_err(anyhow::Error::from)?;
+
+        tx.execute("DELETE FROM files WHERE user_id = $1", &[&user_id])
```
hanakannzashi, line 387: This only deletes local `files` rows. The normal file deletion path deletes the provider/OpenAI file first, then removes the DB row; account deletion skips that upstream cleanup and then loses the provider file ids stored in `files.id`. Please have the account-deletion worker delete provider files before finalization, or otherwise verify upstream file deletion before removing these local rows.

crates/database/src/repositories/user_repository.rs
```diff
@@ -139,16 +227,367 @@ impl UserRepository for PostgresUserRepository {
...
+                 (instance_id, old_status, new_status, changed_by_user_id, change_reason, changed_at)
+             SELECT id, old_status, 'deleted', NULL, 'user_account_deleted', NOW()
+             FROM updated
+             WHERE old_status != 'deleted'",
+            &[&user_id],
+        )
+        .await
+        .map_err(anyhow::Error::from)?;
+
+        let deleted = tx
```
hanakannzashi, line 480: This deletes the `users` row without first removing email-auth challenge rows keyed only by email/IP (`email_verification_challenges`). After the user is gone, those rows still retain PII and can no longer be associated back to a user id for cleanup. Please load the user's email before this point and delete or anonymize matching email verification challenge rows in the same deletion transaction.

crates/api/src/routes/users.rs
```diff
@@ -41,6 +44,101 @@ pub async fn get_current_user(
...
+    );
+
+    let task_publisher = app_state
+        .account_deletion_task_publisher
+        .as_ref()
+        .ok_or_else(|| {
+            ApiError::internal_server_error("Account deletion queue is not configured")
+        })?;
+
+    let deletion = app_state
```
hanakannzashi: Creating and committing the deletion request before publishing the SQS task can leave the account stuck if publish fails. The API returns 500, but auth will now block all non-delete requests because the user has a pending deletion request, and nothing later re-enqueues the task. Please add an outbox/sweeper for pending deletion rows or mark/rollback the request when enqueue fails.

crates/services/src/conversation/service.rs
```diff
@@ -161,6 +161,13 @@ impl ConversationService for ConversationServiceImpl {
 
         Ok(deleted)
     }
+
+    async fn delete_conversation_from_provider(
+        &self,
+        conversation_id: &str,
+    ) -> Result<serde_json::Value, ConversationError> {
```
hanakannzashi, line 168: This account-deletion-specific entry point should treat provider-side "already deleted/not found" responses such as 404 as success so retries are idempotent. Otherwise, if the upstream delete succeeds but the later progress write fails, the next worker retry will attempt to delete the same conversation again and can get stuck on a provider non-200 until the deletion moves to manual review.

crates/api/src/bin/task_worker.rs
```diff
@@ -180,6 +209,157 @@ impl TaskExecutor for DefaultTaskExecutor {
...
+            self.user_repository
+                .mark_account_deletion_failed_needs_review(
```
think-in-universe, line 289: Do we have an admin endpoint to review all the failed account deletion? How will we rerun the deletion for all the failed accounts?

crates/database/src/repositories/user_repository.rs
```diff
@@ -12,6 +18,88 @@ impl PostgresUserRepository {
...
+            return Err(AccountDeletionError::ActiveSubscriptions {
+                count: active_subscription_count,
+            });
+        }
+
+        let blocking_instance_rows = client
+            .query(
+                "SELECT status, COUNT(*)::bigint AS count
+                 FROM agent_instances
+                 WHERE user_id = $1 AND status NOT IN ('stopped', 'deleted')
```
think-in-universe: Should we enforce all agent instances to be deleted, not stopped, before users can delete the account?

crates/database/src/repositories/user_repository.rs
```diff
@@ -139,16 +227,367 @@ impl UserRepository for PostgresUserRepository {
...
+            .map(|row| row.get::<_, String>("id"))
+            .filter(|id| !verified_conversation_ids.contains(id.as_str()))
+            .collect();
+        if !missing_cloud_deletes.is_empty() {
+            return Err(AccountDeletionError::ConversationCleanupIncomplete {
+                conversation_ids: missing_cloud_deletes,
+            });
+        }
+
+        tx.execute(
```
think-in-universe, line 358: Should we add comments about which tables are kept after account deletion?

crates/database/src/repositories/user_repository.rs
danielwpz: Good catch. Added provider file deletion in the worker before the DB cleanup step. File IDs are fetched via list_owned_file_ids and deleted through the provider API, treating 404 as already-deleted.

crates/database/src/repositories/user_repository.rs
danielwpz: Fixed. The user email is now loaded and email_verification_challenges rows are deleted by email before the users row is removed, all within the same transaction.

crates/api/src/routes/users.rs
danielwpz: Fixed. When SQS publish fails, the handler now deletes the just-created user_account_deletions row so the user is not soft-locked and can retry.

crates/services/src/conversation/service.rs
danielwpz: Fixed. delete_conversation_from_openai now treats HTTP 404 as success so retries are idempotent. Same 404-safe handling added for provider file deletion.

crates/api/src/bin/task_worker.rs
danielwpz: Added GET /v1/admin/account-deletions with optional ?status= filter and pagination (limit, offset). Returns full deletion records including status, attempt_count, last_error, progress. When a deletion fails, the worker writes the exact failure reason to last_error with attempt_count tracking — check logs and this endpoint to decide how to proceed.

crates/database/src/repositories/user_repository.rs
danielwpz: Done — added a doc comment on `delete_user_account` listing exactly which tables are retained and why:

- `agent_usage_log`, `user_usage_event`, `credit_transactions` — billing/audit data, user_id FK was intentionally removed in V34 so rows survive the users row deletion.
- `agent_instance_status_history` — `changed_by_user_id` is SET NULL on instance delete; history rows remain.
- `user_account_deletions` — the deletion audit record itself.

crates/api/src/routes/users.rs
```diff
@@ -41,6 +44,120 @@ pub async fn get_current_user(
...
+        ApiError::internal_server_error("Failed to enqueue account deletion")
+    })?;
+    let message = TaskMessage {
+        task_id,
+        payload: TaskPayload::AccountDeletion(AccountDeletionTaskPayload {
+            deletion_id: deletion.id,
+        }),
+    };
+
+    if let Err(e) = task_publisher.publish(message).await {
```
hanakannzashi: This rollback can delete an existing account deletion request. `create_account_deletion_request` returns the row from `ON CONFLICT DO UPDATE`, so a retry of DELETE /users/me can get the already-pending request; if this second publish fails, we delete the original request and unblock the account. Please only rollback rows that were newly inserted by this request, or make create return an inserted/existing flag and skip rollback for existing rows.

crates/api/src/bin/task_worker.rs
```diff
@@ -180,6 +266,192 @@ impl TaskExecutor for DefaultTaskExecutor {
...
+                .update_account_deletion_progress(
+                    request.id,
+                    build_account_deletion_progress(&cloud_deleted_conversation_ids),
+                    ACCOUNT_DELETION_LEASE_SECONDS,
+                )
+                .await
+                .context("failed to update account deletion progress")?;
+        }
+
+        let file_ids = self
```
hanakannzashi, line 361: Files need the same finalization guard as conversations. We list/delete provider files before finalization, but `delete_user_account` later deletes all current `files` rows without verifying that every current file id was deleted from the provider. An in-flight upload that passed auth before the deletion request can be tracked after `list_owned_file_ids`, causing the local row to be removed while the provider file remains. Please track/pass `cloud_deleted_file_ids` and verify the current DB file set before deleting local rows, or otherwise lock out/serialize file creation before listing.

crates/api/src/bin/task_worker.rs
```diff
@@ -180,6 +266,192 @@ impl TaskExecutor for DefaultTaskExecutor {
...
+            .user_repository
+            .claim_account_deletion(payload.deletion_id, ACCOUNT_DELETION_LEASE_SECONDS)
+            .await
+            .context("failed to claim account deletion request")?
+        else {
+            tracing::info!(
+                "account deletion request is currently leased by another worker deletion_id={}",
+                payload.deletion_id
+            );
+            return Ok(());
```
hanakannzashi: Returning Ok here can ack/delete an SQS redelivery while another worker is still processing the same deletion. If that original worker later fails or crashes, the message may already be gone and the deletion request can be left stuck without a retry. Please avoid acknowledging leased account-deletion messages, or keep extending SQS visibility while the owner is still working.

crates/database/src/repositories/user_repository.rs
danielwpz: Fixed — account deletion now requires all user instances to be deleted first.

crates/api/src/routes/users.rs
```diff
@@ -41,6 +44,120 @@ pub async fn get_current_user(
...
+        tracing::error!(
+            "Failed to enqueue account deletion task, cleaning up request: user_id={}, deletion_id={}, error={:#}",
+            user.user_id,
+            deletion.id,
+            e
+        );
+        let _ = app_state
+            .user_service
+            .delete_account_deletion_request(deletion.id)
+            .await;
```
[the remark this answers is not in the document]
danielwpz: Fixed — creation now returns whether the deletion request was newly inserted, and rollback only deletes requests inserted by the current call.

crates/database/src/repositories/user_repository.rs
```diff
@@ -139,16 +227,456 @@ impl UserRepository for PostgresUserRepository {
...
+    /// data, billing rows (stripe_customers, user_credits, subscriptions*),
+    /// sharing groups, agent api keys, activity log, bans, passkey credentials,
+    /// email verification challenges, and the users row itself.
+    /// Agent instance provider deletion is handled by the account-deletion worker
+    /// before this DB finalization step; this function then soft-deletes any
+    /// remaining non-deleted instance rows and wipes stored credentials.
+    ///
+    /// Intentionally retained for audit / billing reconciliation:
+    /// - agent_usage_log, user_usage_event, credit_transactions – their user_id
+    ///   FK was removed in V34 so they survive the users row deletion.
```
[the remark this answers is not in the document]
danielwpz: Fixed — subscriptions are now documented as intentionally retained audit/billing data instead of deleted account data.

crates/database/src/repositories/user_repository.rs
```diff
@@ -12,6 +18,88 @@ impl PostgresUserRepository {
...
+            .is_some();
+        if !user_exists {
+            return Err(AccountDeletionError::UserNotFound);
+        }
+
+        let active_subscription_count: i64 = client
+            .query_one(
+                "SELECT COUNT(*)
+                 FROM subscriptions
+                 WHERE user_id = $1 AND status IN ('active', 'trialing')",
```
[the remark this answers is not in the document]
danielwpz: Fixed — subscription upserts now take a user row key-share lock before writing, so they serialize with account deletion’s user row lock and fail if the user has already been deleted.

crates/database/src/repositories/user_repository.rs
```diff
@@ -139,16 +227,456 @@ impl UserRepository for PostgresUserRepository {
...
+            .collect();
+        if !missing_cloud_deletes.is_empty() {
+            return Err(AccountDeletionError::ConversationCleanupIncomplete {
+                conversation_ids: missing_cloud_deletes,
+            });
+        }
+
+        tx.execute(
+            "DELETE FROM conversation_shares
+             WHERE owner_user_id = $1
```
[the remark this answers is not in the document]
danielwpz: Fixed — account deletion now removes direct shares where the deleted user’s email or linked NEAR account is the recipient.

crates/database/src/repositories/user_repository.rs
```diff
@@ -139,16 +227,456 @@ impl UserRepository for PostgresUserRepository {
...
+                OR conversation_id IN (SELECT id FROM conversations WHERE user_id = $1)
+                OR group_id IN (SELECT id FROM conversation_share_groups WHERE owner_user_id = $1)",
+            &[&user_id],
+        )
+        .await
+        .map_err(anyhow::Error::from)?;
+
+        tx.execute(
+            "DELETE FROM conversation_share_group_members
+             WHERE group_id IN (SELECT id FROM conversation_share_groups WHERE owner_user_id = $1)",
```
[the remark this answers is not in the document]
danielwpz: Fixed — account deletion now also removes group members matching the deleted user’s email or linked NEAR account.

crates/api/src/routes/admin.rs
```diff
@@ -3017,6 +3017,54 @@ pub async fn bi_top_consumers(
...
+        Some(s) => Some(
+            services::user::ports::AccountDeletionStatus::parse_db_value(s)
+                .map_err(|_| ApiError::bad_request(format!("invalid status filter: {s}")))?,
+        ),
+    };
+
+    let limit = params.limit.min(200);
+    let deletions = app_state
+        .user_service
+        .list_account_deletions(status, limit, params.offset)
```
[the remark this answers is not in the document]
danielwpz: Fixed — admin deletion listing now clamps limit to 1..=200 and offset to >= 0.

crates/api/src/routes/users.rs
danielwpz: Fixed — rollback is now limited to deletion requests newly inserted by the current DELETE call, so retries cannot delete an existing pending request.

crates/api/src/bin/task_worker.rs
danielwpz: Fixed — files now use the same finalization guard as conversations: provider-deleted file ids are tracked in progress and verified before local file rows are removed.

crates/api/src/bin/task_worker.rs
danielwpz: Fixed — a leased account-deletion redelivery now returns an error instead of success, so the SQS message is not acknowledged while another worker owns the lease.

think-in-universe (approved):

think-in-universe (approved):

PR #275, opened 2026-05-04
feat: payment by staking and paying near to House of Stake contract

think-in-universe:
## Summary

Adds **House of Stake (HoS)** as an optional subscription billing path alongside Stripe. Users who sign in with a **linked NEAR account** can subscribe using the on-chain staking contract; chat-api returns **client/wallet intents** (catalog `price_id`, network id, upgrade/downgrade/cancel/resume guidance) and keeps a local `house-of-stake` subscription row aligned with chain state via **NEAR JSON-RPC view calls** and **`POST /v1/subscriptions/near/sync`**.

## Motivation

Support subscription access tied to **staked NEAR / House of Stake** without duplicating payment logic in the API: wallets execute contract calls; the backend reconciles view state into the existing subscriptions model for quotas and product behavior.

## What changed

- **Config**: `NEAR_STAKING_CONTRACT_ID`, `NEAR_RPC_URL`, `NEAR_NETWORK_ID`; optional `near.staking_contract_id` in TOML (legacy alias `near_staking_contract_id`).
- **New** `subscription::near_staking`: view calls `get_subscription_for_price`, `get_price`; map chain JSON → DB row; 15s RPC timeouts.
- **Subscription service**: `provider: "house-of-stake"` on create (returns `kind` + `price_id` + `network_id`), cancel/resume/change_plan HoS branches, reconcile upsert/delete, **guard** so HoS upsert does not override an active **non-HoS** subscription (`skipped_reason` on sync).
- **API**: OpenAPI updates; `POST /v1/subscriptions/near/sync`; admin error mapping for missing NEAR wallet where relevant.
- **Tests**: API coverage for HoS create, wallet requirement, cancel/resume, sync edge cases, delete when chain returns null, upgrade shape, Stripe blocking HoS upsert.

## How to test

```bash
cargo test -p services subscription::near_staking
cargo test --test subscriptions_tests --features test
```

Manual smoke (with HoS configured):

1. Link a NEAR wallet to a user; `POST /v1/subscriptions` with `provider: "house-of-stake"` and a valid plan → response includes `price_id` and `network_id`.
2. After on-chain lock/subscribe, call `POST /v1/subscriptions/near/sync` → local `house-of-stake` row upserted; `skipped` / `skipped_reason` behave as documented when Stripe is active.

## Rollout / ops notes

- Set **`NEAR_STAKING_CONTRACT_ID`** to the deployed staking contract account; ensure **`NEAR_RPC_URL`** and **`NEAR_NETWORK_ID`** match the environment (e.g. testnet vs mainnet).
- System config must define per-plan **`providers.house-of-stake.price_id`** (and usual plan limits) so catalog and chain probes stay consistent.


crates/services/src/subscription/service.rs
```diff
@@ -1976,17 +2460,37 @@ impl SubscriptionService for SubscriptionServiceImpl {
...
-        // When Stripe is not configured, allow access (no subscription gating)
-        match self.get_plans_for_provider("stripe").await {
-            Err(SubscriptionError::NotConfigured) => {
-                tracing::debug!(
-                    "Subscription gating skipped: Stripe not configured, allowing user_id={}",
-                    user_id
-                );
-                return Ok(());
+        let stripe_res = self.get_plans_for_provider("stripe").await;
+        let hos_res = self.get_plans_for_provider("house-of-stake").await;
```
hanakannzashi, line 3059: This adds HoS as a configured billing path for proxy gating, but the credit-limit calculation later in this method still resolves every active subscription with `resolve_plan_name_from_config("stripe", &effective_sub.price_id, ...)`. For an active `house-of-stake` row, the HoS price id will not match the Stripe provider config and the user falls back to `DEFAULT_MONTHLY_CREDITS_NANO_USD` instead of the actual HoS plan limit. Please resolve entitlements with the subscription's actual provider and avoid running Stripe-only pending-downgrade handling for HoS rows.

crates/services/src/subscription/near_staking.rs
```diff
@@ -0,0 +1,238 @@
...
+        .ok_or_else(|| "missing price_id".to_string())?
+        .to_string();
+
+    let end_ns = json_u64(
+        v.get("end_ns")
+            .ok_or_else(|| "missing end_ns".to_string())?,
+    )?;
+    let current_period_end = ts_ns_to_datetime(end_ns)?;
+
+    let status_raw = v.get("status").and_then(|x| x.as_str()).unwrap_or("Active");
```
hanakannzashi: This maps an on-chain `Active` HoS subscription to a local active row without checking whether `end_ns` is already in the past. `get_active_subscription` only filters by status, so an expired HoS billing period can continue granting access locally. Please mark the synced row canceled/expired when `current_period_end <= Utc::now()`, or make active-subscription selection enforce the period end for HoS rows.

crates/services/src/subscription/service.rs
```diff
@@ -733,6 +796,234 @@ impl SubscriptionServiceImpl {
...
+            .db_pool
+            .get()
+            .await
+            .map_err(|e| SubscriptionError::DatabaseError(e.to_string()))?;
+        let txn = db_client
+            .transaction()
+            .await
+            .map_err(|e| SubscriptionError::DatabaseError(e.to_string()))?;
+
+        self.subscription_repo
```
hanakannzashi, line 1508: The chain sync upserts the HoS row through the generic `upsert_subscription`, but that repository preserves existing pending-downgrade fields whenever the incoming row has `pending_downgrade_status = NULL`. For HoS sync, the chain row should be authoritative: if `pending_downgrade_price_id` is absent on-chain, local pending-downgrade state should be cleared. Otherwise the API can keep showing or acting on a stale pending downgrade after the chain has cleared it.

crates/services/src/subscription/service.rs
```diff
@@ -733,6 +799,288 @@ impl SubscriptionServiceImpl {
...
+        // subscriptions for other products.
+        let local_probe_price_id = subs
+            .iter()
+            .filter(|s| s.provider == "house-of-stake")
+            .find(|s| s.status == "active" || s.status == "trialing")
+            .or_else(|| subs.iter().find(|s| s.provider == "house-of-stake"))
+            .map(|s| s.price_id.as_str())
+            .filter(|p| !p.is_empty());
+
+        let raw = if let Some(probe_price_id) = local_probe_price_id {
```
think-in-universe: This only probes the locally stored HoS `price_id` once a HoS row exists. After a successful cross-product HoS plan change, the local row can still contain the old price until sync; if `get_subscription_for_price(old_price)` returns `null`, the code below deletes all local HoS rows even though the user may now have a valid subscription under the new product/price. Before deleting, this should fall back to probing all configured HoS price IDs, or otherwise query by a stable subscription/product identifier.

crates/api/src/routes/agents.rs
```diff
@@ -56,6 +56,18 @@ async fn get_instance_limit(
     app_state: &AppState,
     user_id: services::UserId,
 ) -> Result<(u64, u64), ApiError> {
+    if let Err(e) = app_state
```
think-in-universe: This adds a NEAR RPC sync to every instance-limit check. For NEAR-linked users with HoS configured and no local HoS row, sync can probe every configured HoS price, each with the 15s timeout in `near_staking.rs`. A degraded NEAR RPC can therefore make unrelated agent operations hang for a long time. Consider limiting this to explicit subscription/sync flows, caching the result, or moving reconciliation to a background path.

think-in-universe (commented):
I found two issues worth addressing before merge. The first is a correctness issue in HoS sync after cross-product plan changes; the second is a latency/regression risk from doing RPC reconciliation in agent instance-limit checks.

crates/services/src/subscription/service.rs
```diff
@@ -733,6 +898,317 @@ impl SubscriptionServiceImpl {
...
+            .db_pool
+            .get()
+            .await
+            .map_err(|e| SubscriptionError::DatabaseError(e.to_string()))?;
+        let txn = db_client
+            .transaction()
+            .await
+            .map_err(|e| SubscriptionError::DatabaseError(e.to_string()))?;
+
+        self.subscription_repo
```
hanakannzashi, line 1508: HoS sync upserts the chain subscription by subscription_id but does not remove stale local HoS rows when the on-chain subscription id changes. Cross-product HoS plan changes can leave multiple active house-of-stake rows for the same user. Please delete or deactivate stale local HoS rows that do not match the current chain subscription id in the same transaction.

crates/services/src/subscription/service.rs
```diff
@@ -1282,6 +1846,176 @@ impl SubscriptionService for SubscriptionServiceImpl {
...
+                        .db_pool
+                        .get()
+                        .await
+                        .map_err(|e| SubscriptionError::DatabaseError(e.to_string()))?;
+                    let txn = client
+                        .transaction()
+                        .await
+                        .map_err(|e| SubscriptionError::DatabaseError(e.to_string()))?;
+                    self.subscription_repo
+                        .clear_pending_downgrade(&txn, &subscription.subscription_id)
```
hanakannzashi: This clears the pending downgrade only in the local DB, but the pending downgrade state comes from the HoS contract and will be restored on the next sync. Please return a wallet intent that clears the pending downgrade on-chain, or make this operation explicitly unsupported for HoS subscriptions.

crates/services/src/subscription/service.rs
```diff
@@ -2491,63 +3250,218 @@ impl SubscriptionService for SubscriptionServiceImpl {
...
+                let _near_account = self.get_near_account_id(user_id).await?;
+
+                tracing::info!(
+                    "House-of-Stake credit payment intent created: user_id={}, credits={}, price_id={}",
+                    user_id,
+                    credits,
+                    price_id
+                );
+
+                Ok(CreateCreditPurchaseOutcome::HouseOfStake {
```
hanakannzashi, line 3673: This returns a HoS pay intent before validating that the configured credit price is active and one-off. The confirmation path rejects inactive or non-one-off prices only after the user may have paid. Please run the same get_price validation before returning the wallet intent.

hanakannzashi (commented):
Review comments from Codex.

crates/services/src/subscription/near_staking.rs
```diff
@@ -0,0 +1,328 @@
...
+
+    let cancel_at_period_end = v
+        .get("cancel_at_period_end")
+        .and_then(|x| x.as_bool())
+        .unwrap_or(false);
+    if status == "active" && current_period_end <= Utc::now() {
+        status = "canceled".to_string();
+    }
+
+    let pending_down = v.get("pending_downgrade_price_id").and_then(|x| {
```
danielwpz: Looks like the contract now returns pending changes as `pending_update`, so reading `pending_downgrade_price_id` here may fail now

crates/services/src/subscription/service.rs
```diff
@@ -1249,37 +2016,198 @@ impl SubscriptionService for SubscriptionServiceImpl {
...
+                std::cmp::Ordering::Greater => {
+                    if !same_price && new_amt <= cur_amt {
+                        return Err(SubscriptionError::InvalidTargetAmount(
+                            "target_amount increase requires a higher target price amount".into(),
+                        ));
+                    }
+                    "immediate"
+                }
+                std::cmp::Ordering::Less => {
+                    if !same_price && new_amt >= cur_amt {
```
hanakannzashi: The contract allows plan direction and stake direction to move independently. For example, an upgrade can apply immediately while a lower target_amount is scheduled for period end when the current stake still supports the upgraded price. This branch rejects that valid case whenever target_amount decreases and the target price amount is not lower. Please mirror the contract decision rules, or return the raw wallet intent and let the contract decide, instead of deriving timing from price amount direction alone.

crates/services/src/subscription/ports.rs
```diff
@@ -132,6 +135,200 @@ pub enum ChangePlanOutcome {
     NoOp,
     /// A pending downgrade was cancelled (same plan requested with active pending downgrade).
     DowngradeCancelled,
+    /// HoS: call `update_subscription` in the wallet with this subscription id, price id, and stake amount.
+    NearStakingChangePlan {
```
hanakannzashi, line 139: This wallet intent does not include the NEAR deposit required by update_subscription. The contract requires the attached deposit to equal the stake increase delta, or exactly 1 yoctoNEAR for non-increase updates. Since the service already fetched the current lock amount, please include required_deposit_yocto so clients do not accidentally attach target_amount or omit the 1 yocto deposit.

crates/services/src/subscription/service.rs
```diff
@@ -2491,63 +3460,231 @@ impl SubscriptionService for SubscriptionServiceImpl {
...
+                "House-of-Stake purchase is missing quantity".to_string(),
+            )
+        })?;
+        if quantity != expected_credits {
+            return Err(SubscriptionError::InvalidCredits(
+                "House-of-Stake purchase quantity does not match requested credits".to_string(),
+            ));
+        }
+
+        if !Self::price_is_active(&price) || !Self::price_is_one_off(&price) {
```
hanakannzashi: Confirmation should not require the current catalog price to still be Active. The contract's pay method already enforced active one-off and exact amount before recording the Purchase. If the price is archived after payment but before confirm, this check rejects a legitimate paid purchase. Please validate the immutable purchase fields and amount/quantity, and move active-price validation to checkout creation if needed for UX.

hanakannzashi (commented):
Reviewed the latest HoS payment/subscription changes against the staking contract behavior.

crates/services/src/subscription/service.rs
```diff
@@ -2057,22 +2977,38 @@ impl SubscriptionService for SubscriptionServiceImpl {
...
                         );
-                        let plan_credits = plan_name
-                            .as_deref()
-                            .and_then(|n| subscription_plans.get(n))
-                            .map(Self::plan_limit_max)
-                            .unwrap_or_else(|| {
+                        let plan_credits = match plan_name.as_deref() {
+                            Some(name) => match subscription_plans.get(name) {
+                                Some(plan_config) => self
+                                    .stake_based_plan_limit_max(
```
hanakannzashi, line 3167: [P1] This entitlement path can continue granting HoS access from a stale local row after the on-chain subscription has ended. get_active_subscription only filters status IN ('active', 'trialing'), and the proxy/model paths do not force a HoS reconcile or require current_period_end > now before using the row. A user can sync once, cancel on-chain, wait for the period to end, unlock/withdraw the NEAR, and still be treated as paid as long as the local row remains active. Even this stake-based credit path only reads last_lock_id/amount_near and does not verify the chain subscription is still active or the lock is still active. Please expire stale rows locally and/or refresh HoS state before granting paid proxy/model access.

hanakannzashi (commented):
Additional HoS entitlement review against the staking contract lifecycle.

danielwpz (approved):

hanakannzashi (approved):

PR #280, opened 2026-05-14
fix: add cache read tokens usage tracking and cache-aware billing

hanakannzashi:
## Summary

Fix chat-api usage tracking for cached input tokens returned by cloud-api.

- Parse cached tokens from upstream usage payloads:
  - Chat Completions: `usage.prompt_tokens_details.cached_tokens`
  - Responses API: `usage.input_tokens_details.cached_tokens`
- Store `cached_tokens` in `user_usage_event.details` alongside existing `input_tokens`, `output_tokens`, and cost breakdown fields.
- Apply cache-read pricing from cloud-api model pricing via `cacheReadCostPerToken`.
- Keep legacy behavior when cache-read pricing is missing or zero: cached tokens are billed at the normal input-token rate.
- Expose `cached_tokens` in agent instance usage detail responses.

No DB migration is required because detailed token breakdowns are already stored in the existing `details` JSONB column.

## Testing

- `cargo fmt`
- `cargo test -p api usage_parsing`
- `cargo test -p api model_pricing`


crates/api/src/usage_parsing.rs
```diff
@@ -23,6 +25,7 @@ pub struct ParsedUsage {
     pub input_tokens: u64,
     pub output_tokens: u64,
     pub total_tokens: u64,
+    pub cached_tokens: u64,
```
think-in-universe: Should we use `cache_read_tokens` to avoid conflicts with potential `cache_write_tokens`, etc.?

think-in-universe (approved):

PR #287, opened 2026-05-27
feat: add explicit admin email allowlist and enforce in admin middleware

think-in-universe:
### Motivation
- Introduce an explicit allowlist of admin email addresses in addition to domain-based checks to tighten admin access control.
- Provide test utilities to configure and validate email-level admin allowlisting in tests.

### Description
- Add `admin_emails: Vec<String>` to `AdminConfig` and parse `AUTH_ADMIN_EMAILS` in `config::AdminConfig::default` while filtering invalid entries and those outside `AUTH_ADMIN_DOMAINS`.
- Add `admin_emails` to `AppState` and `AuthState`, propagate it through router creation in `create_router_with_cors`, and populate it in `main.rs` when building `AppState`.
- Implement `is_admin_email` helper and update `admin_auth_middleware` to require both domain membership and explicit email allowlist via `is_admin_domain` and `is_admin_email` checks.
- Extend test support: add `admin_emails` to `TestServerConfig`, wire test overrides in `create_test_server_and_db`, and add tests `test_admin_users_list_denied_when_email_not_allowlisted` and `test_admin_users_list_allowed_when_email_allowlisted` in `crates/api/tests/admin_tests.rs`.
- Update `env.example` to include `AUTH_ADMIN_EMAILS`.

### Testing
- Ran the API integration tests including the new admin allowlist tests under `crates/api/tests/*`, and the suite passed locally (new tests `test_admin_users_list_denied_when_email_not_allowlisted` and `test_admin_users_list_allowed_when_email_allowlisted` exercised the change).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a169c58133483309b6aa9472830e126)

### DIFF
--- a/crates/api/tests/common.rs
+++ b/crates/api/tests/common.rs
@@ -43,6 +43,10 @@ pub struct TestServerConfig {
     pub email_turnstile_verify_url: Option<String>,
     /// Optional override to enable/disable email auth in tests.
     pub email_auth_enabled: Option<bool>,
+    /// Optional override for admin domains allowlist in tests.
+    pub admin_domains: Option<Vec<String>>,
+    /// Optional override for explicit admin emails allowlist in tests.
+    pub admin_emails: Option<Vec<String>>,
 }
 
 /// Restrictive rate limit config for rate limit tests.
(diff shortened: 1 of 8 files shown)


crates/api/tests/common.rs
```diff
@@ -257,7 +261,14 @@ pub async fn create_test_server_and_db(
         user_repo.clone(),
     ));
 
-    let mut admin_domains = config.admin.admin_domains;
+    let mut admin_domains = test_config
+        .admin_domains
+        .clone()
+        .unwrap_or(config.admin.admin_domains);
+    let admin_emails = test_config
```
hanakannzashi: Could we default test `admin_emails` to an empty Vec here instead of inheriting `config.admin.admin_emails`? This shared test helper loads the real `.env` / environment, so if a developer or CI has `AUTH_ADMIN_EMAILS` set, legacy admin tests that rely on the injected `admin.org` domain will unexpectedly enforce that real email allowlist and fail with 403.

Repro:
`AUTH_ADMIN_DOMAINS=admin.org AUTH_ADMIN_EMAILS=someone_else@admin.org cargo test -p api --features test --test admin_tests test_admin_users_list_with_admin_account -- --exact --nocapture`

The allowlist-specific tests already pass `admin_emails` explicitly, so the default helper should stay domain-only/hermetic, e.g. `unwrap_or_default()`.

hanakannzashi (approved):

PR #288, opened 2026-05-27
fix: oauth frontend callback validation

think-in-universe:
## Summary

Fixes an open redirect risk in the OAuth flow by validating `frontend_callback` before it is stored in OAuth state and before it is used as a redirect destination.

The redirect validation is kept local to `oauth.rs` and now:
- requires `frontend_callback` for OAuth login initiation
- requires the callback origin to match `FRONTEND_CALLBACK_ALLOWED_ORIGINS`
- rejects malformed URLs, non-HTTP(S) schemes, credentials, query strings, fragments, empty allowlists, and untrusted origins
- requires HTTPS for remote origins while still allowing loopback HTTP for local development
- builds the final `/auth/callback` redirect with `url::Url` instead of string concatenation

## Impact

Clients initiating Google/GitHub OAuth must provide a trusted `frontend_callback`. Missing or invalid callbacks now fail closed instead of falling back to a server-side frontend URL.

## Validation

- `cargo test -p api --lib routes::oauth::tests`
- `cargo check -p api --bins`

### DIFF
--- a/crates/api/src/routes/oauth.rs
+++ b/crates/api/src/routes/oauth.rs
@@ -18,9 +18,41 @@ use services::metrics::consts::{
     METRIC_USER_LOGIN, METRIC_USER_SIGNUP, TAG_AUTH_METHOD, TAG_IS_NEW_USER,
 };
 use services::SessionId;
+use std::collections::HashSet;
 use std::convert::Infallible;
 use std::net::{IpAddr, SocketAddr};
+use url::{Host, Url};
 use utoipa::ToSchema;
+
+const FRONTEND_CALLBACK_ALLOWED_ORIGINS_ENV: &str = "FRONTEND_CALLBACK_ALLOWED_ORIGINS";
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum FrontendCallbackValidationError {
+    Malformed,
+    UnsupportedScheme,
+    MissingHost,
+    CredentialsNotAllowed,
+    QueryOrFragmentNotAllowed,
+    Insecure,
+    UntrustedOrigin,
+}
+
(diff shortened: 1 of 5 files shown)


hanakannzashi (approved):

crates/api/src/routes/oauth.rs
```diff
@@ -61,6 +93,152 @@ fn try_add_gateway_cookie(headers: &mut HeaderMap, cookie: Option<String>, conte
...
+    let mut origins = Vec::new();
+    let mut seen = HashSet::new();
+
+    for origin in raw_origins
+        .split(',')
+        .map(str::trim)
+        .filter(|s| !s.is_empty())
+    {
+        match validate_frontend_origin(origin) {
+            Ok(url) => insert_origin(&mut origins, &mut seen, &url),
```
walnut-the-cat, line 115: nit: it seems unnecessary to have both origins and seen, if it's purely for dedupe?

crates/api/src/routes/oauth.rs
```diff
@@ -61,6 +93,152 @@ fn try_add_gateway_cookie(headers: &mut HeaderMap, cookie: Option<String>, conte
...
+    allowed_origins: Option<&[String]>,
+) -> Result<String, FrontendCallbackValidationError> {
+    if let Some(origins) = allowed_origins {
+        if origins.is_empty() {
+            return Err(FrontendCallbackValidationError::UntrustedOrigin);
+        }
+    }
+
+    validate_frontend_url(raw_url, allowed_origins.unwrap_or(&[]))
+        .map(|url| trim_trailing_slash(url.as_str()))
```
walnut-the-cat, line 136: nit: no need to create trim_trailing_slash if it's used only here

PR #300, opened 2026-06-08
Fix migration SSE timeout and add step-level progress logging

Longarithm:
## Summary
- Rename `SSE_CHUNK_TIMEOUT` (30s) → `SSE_MIGRATION_CHUNK_TIMEOUT` (600s) for backup and import SSE parsers
- Add `tracing::info` at every migration step with elapsed time and instance_id
- Add `tracing::info` for every SSE stage transition in both backup and import parsers
- Include chunk count and elapsed time in timeout/EOF error messages
- Catch `stage=complete` with null `download_url` explicitly

## Problem
Every migration attempt fails with: `Timed out waiting for SSE chunk from backup stream (30s)`.

The migration endpoint POSTs to the legacy compose-api backup endpoint, which returns an SSE stream. The backup process (docker exec tar + age encrypt + S3 upload) takes ~52s for a 298MB full export. During that time, the SSE stream has no data events — only keepalive comments, which arrive late (~43s instead of the configured 15s). The 30s per-chunk timeout fires before any data arrives.

## Investigation

1. **Confirmed the error** via Datadog logs on `cpu01-chat-api-staging`: `Failed to parse backup SSE stream: instance_id=159407f7..., error=Timed out waiting for SSE chunk from backup stream (30s)`.

2. **Verified the backup endpoint works** — curled directly from the legacy host (`cpu03`), SSE stream completes successfully: `encrypting` → keepalives → `uploading` → `complete` with `download_url`. Backup produces 298MB in ~15s from host.

3. **Checked compose-api logs** — zero on-demand backup requests from chat-api migration attempts. Only scheduled backups visible. Initially suspected networking/VPC issue between chat-api CVM (cpu01) and legacy TEE CVM (cpu03).

4. **Reproduced with Python `http.client`** against the compose-api directly (port 10008, bypassing nginx entirely): same failure — first SSE event arrives, then timeout. Ruled out nginx buffering.

5. **Tested with 120s timeout** against direct compose-api port: first event at 0.0s, first keepalive at 43.5s, backup complete at 52.7s. Data DOES arrive — just slower than 30s.

6. **Built minimal axum 0.8 reproduction** with `spawn_blocking` + 5s keepalive: Python client receives keepalives on time at 5s intervals. The 43s delay is specific to the production compose-api deployment (axum 0.8 keepalive timer jitter under load — not fully understood).

7. **Root cause**: `SSE_CHUNK_TIMEOUT` (30s) < backup duration (~52s) + delayed keepalives (~43s). The timeout fires before any keepalive or completion event arrives.

## Fix
- `SSE_MIGRATION_CHUNK_TIMEOUT` = 600s (matches the HTTP request timeout). Only catches truly dead connections.
- Step-level logging so next time we can see exactly where time is spent without needing to reproduce.

## Test plan
- [ ] Deploy to staging
- [ ] Trigger migration on mighty-jay (instance `159407f7`) — should complete
- [ ] Verify Datadog shows step-by-step progress logs
- [ ] Verify the full migration flow (backup → stop → import → DB update)

### DIFF
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -3349,7 +3349,9 @@ pub async fn admin_migrate_instance(
     let id = Uuid::parse_str(&instance_id)
         .map_err(|_| ApiError::bad_request("Invalid instance ID format"))?;
 
-    tracing::info!("Admin: Starting migration for instance_id={}", id);
+    let migrate_start = std::time::Instant::now();
+    let id_str = id.to_string();
+    tracing::info!("Migrate: starting, instance_id={}", id);
 
     let instance = app_state
         .agent_repository
@@ -3442,7 +3444,12 @@ pub async fn admin_migrate_instance(
         ));
     }
 
-    // Step 5: GET instance metadata from compose-api
+    tracing::info!(
+        "Migrate: preflight passed, fetching compose data, elapsed={:.1}s, instance_id={}, name={}",
+        migrate_start.elapsed().as_secs_f64(),
+        id,
+        instance.name,
+    );
     let compose_api_url = agent_api_base_url.trim_end_matches('/');
     let encoded_name = urlencoding::encode(&instance.name);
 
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
Inspected the admin migration diff in crates/api/src/routes/admin.rs, including the SSE backup/import parser changes and surrounding migration flow. I did not find a substantive correctness or operational issue in this patch.\n\nLocal checks run:\n- cargo fmt --check\n- cargo check -p api --features test\n- cargo clippy -p api --all-targets --features test -- -D warnings\n\nGitHub Test Suite was still pending when reviewed.

PR #301, opened 2026-06-08
fix(api): machine-readable code for plan-gated model errors

Evrard-Nil:
## Problem

When a user in Private Chat picks a model that isn't in their subscription plan's `allowed_models`, `enforce_model_access` returns:

```
403 {"error":"Model 'deepseek-ai/DeepSeek-V4-Flash' is not available in your plan 'starter'"}
```

The body carries only a human-readable `error` string, so clients can't distinguish a **gated-model** rejection from a generic upstream failure. The private-chat frontend therefore renders a misleading **"Model … failed to respond"** toast (it should prompt the user to upgrade).

## Change

Add a stable, machine-readable `code` (plus `model` and `plan`) to the 403 body on the `ModelNotAllowedInPlan` path:

```json
{
  "error": "Model 'deepseek-ai/DeepSeek-V4-Flash' is not available in your plan 'starter'",
  "code": "model_not_allowed_in_plan",
  "model": "deepseek-ai/DeepSeek-V4-Flash",
  "plan": "starter"
}
```

- `error` is **byte-for-byte unchanged** → fully backward compatible (existing tests assert only `body["error"]`).
- One call site (`enforce_model_access`), which covers `/v1/chat/completions`, `/v1/responses`, and image endpoints.

## Frontend counterpart

nearai/private-chat#353 consumes `code` to show an "Upgrade to access this model" dialog instead of the generic toast.

## Testing

- `cargo check -p api` ✅
- Existing `model_allowlist_tests` assert `body["error"]` only → unaffected.

### DIFF
--- a/crates/api/src/openapi.rs
+++ b/crates/api/src/openapi.rs
@@ -166,6 +166,7 @@ use utoipa::OpenApi;
         // Conversation share models
         crate::routes::api::ErrorResponse,
         crate::routes::api::ModelNotAllowedErrorResponse,
+        crate::routes::api::ForbiddenErrorResponse,
         crate::routes::api::ShareRecipientPayload,
         crate::routes::api::ShareTargetPayload,
         crate::routes::api::CreateConversationShareRequest,
--- a/crates/api/src/openapi.rs
+++ b/crates/api/src/openapi.rs
@@ -165,6 +165,7 @@ use utoipa::OpenApi;
         crate::routes::admin::TopUsageResponse,
         // Conversation share models
         crate::routes::api::ErrorResponse,
+        crate::routes::api::ModelNotAllowedErrorResponse,
         crate::routes::api::ShareRecipientPayload,
         crate::routes::api::ShareTargetPayload,
         crate::routes::api::CreateConversationShareRequest,
--- a/crates/api/src/routes/api.rs
+++ b/crates/api/src/routes/api.rs
@@ -254,6 +254,23 @@ pub struct ModelNotAllowedErrorResponse {
     pub plan: String,
 }
 
+/// `403` body for the inference/proxy endpoints. It is one of two shapes:
+/// a plan-gated rejection ([`ModelNotAllowedErrorResponse`], carrying
+/// `code = "model_not_allowed_in_plan"`), or a generic forbidden error
+/// ([`ErrorResponse`], e.g. a banned user) that exposes only `error`.
+/// Modeled as a `oneOf` so generated clients accept both bodies.
+///
+/// Documentation-only: handlers return the concrete `ModelNotAllowedErrorResponse`
+/// / `ErrorResponse` directly, so the variants are never constructed in code —
+/// this type exists purely to shape the OpenAPI `403` response.
+#[derive(Serialize, ToSchema)]
+#[serde(untagged)]
+#[allow(dead_code)]
+pub enum ForbiddenErrorResponse {
+    ModelNotAllowed(ModelNotAllowedErrorResponse),
+    Generic(ErrorResponse),
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 struct InvalidProxyPathSegment;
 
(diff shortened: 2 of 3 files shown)


PierreLeGuen (changes requested):
Two issues to address before this is ready:

- crates/api/src/routes/api.rs:3352 adds the new machine-readable payload, but no test asserts the actual new contract. Existing `model_allowlist_tests` only check status and/or the legacy `error` string, so deleting `code`, `model`, or `plan` would still pass. Please extend at least one blocked allowlist case to assert `code == "model_not_allowed_in_plan"` plus the echoed `model` and `plan` values.

- crates/api/src/routes/api.rs:2289, crates/api/src/routes/api.rs:3875, crates/api/src/routes/api.rs:4027, and crates/api/src/routes/api.rs:4222 still document the relevant 403s as `ErrorResponse`, which only exposes `error`. Since this PR is creating a stable client-facing contract consumed by another frontend PR, the OpenAPI schema/docs should describe the extra fields for these plan-gated 403 responses.

Local checks: `cargo check -p api` passed. Focused `cargo test -p api --test model_allowlist_tests test_subscription_plan_allowed_models_blocks_unlisted_model -- --exact --nocapture` built but could not run locally because the test DB connection was refused during migrations. GitHub `Test Suite` was still pending when reviewed.

PierreLeGuen (changes requested):
Requesting changes for one API contract gap.

- `crates/api/src/routes/api.rs:2289`, `crates/api/src/routes/api.rs:3871`, `crates/api/src/routes/api.rs:4023`, `crates/api/src/routes/api.rs:4218`: the shared `enforce_model_access` path now returns a stable plan-gated 403 payload with `code`, `model`, and `plan`, but the affected endpoint OpenAPI annotations still declare the 403 body as `ErrorResponse`, whose schema only has `error` (`crates/api/src/routes/api.rs:234`). Generated clients and Swagger docs will not see the new machine-readable contract. Please add/document a schema for this response shape, or use an appropriate alternative/oneOf response for the plan-gated 403s, so the published API contract matches runtime behavior.

Checks run locally on PR head `4ac2438`:
- `cargo fmt --all -- --check` passed
- `cargo clippy -p api --tests --features test -- -D warnings` passed
- `cargo test -p api --features test test_responses_endpoint_respects_model_allowlist` compiled, then failed before assertions because local Postgres was unavailable (`Connection refused` during migrations)

PierreLeGuen (changes requested):
Requesting changes for one API-contract issue.

- `crates/api/src/routes/api.rs:2308` (also `:3895`, `:4047`, `:4242`): these annotations now document every 403 as `ModelNotAllowedErrorResponse`, whose `code`, `model`, and `plan` fields are required. But these same endpoints call `ensure_user_not_banned` before model access, and that path still returns `ErrorResponse` with only `error` (`crates/api/src/routes/api.rs:2830`). Generated clients/Swagger will now reject or misrepresent valid banned-user 403s. Please model the 403 response as a union/oneOf, or keep the endpoint 403 as `ErrorResponse` and separately document the plan-gated shape, so both actual 403 bodies are in the contract.

Checks: `cargo fmt --check` passed. `cargo test -p api test_responses_endpoint_respects_model_allowlist --test model_allowlist_tests` compiled but could not run locally because migrations could not connect to a database (`Connection refused`). GitHub Test Suite was still pending when reviewed.

PierreLeGuen (approved):
Reviewed the runtime change in `enforce_model_access`, the updated OpenAPI 403 schema, and the affected callers for `/v1/responses`, `/v1/chat/completions`, `/v1/images/generations`, and `/v1/images/edits`. I did not find substantive issues.

Local checks:
- `cargo check -p api` passed.
- `cargo test --lib --bins --features test` passed.
- `cargo test --test model_allowlist_tests --features test` compiled but could not run in this environment because local PostgreSQL was unavailable; all tests failed during migrations at `crates/api/tests/common.rs:120` with connection refused.

lloydmak99 (approved):
Reviewed — approve. Clean, backward-compatible addition of code/model/plan fields to the 403 body. Test assertions cover the exact frontend contract. ForbiddenErrorResponse oneOf union is the right OpenAPI shape.

PR #302, opened 2026-06-08
feat: allow validated mobile auth callbacks

abbyshekit:
## Summary
- Allow exact mobile OAuth `frontend_callback` URLs for the current iOS/Android scheme `nearprivatechat://auth`, the new shared scheme `privatechat://auth`, and the universal link `https://app.privatechat.com/auth/callback`.
- Preserve the mobile `state` query while appending session token fields, so the app can complete its active pending sign-in instead of opening into a state-validation failure.
- Avoid appending `/auth/callback` onto custom-scheme or already-exact universal-link callbacks.
- Update CORS/callback environment examples and the default CORS origin list for `https://app.privatechat.com`.

## Validation
- `cargo fmt`
- `cargo test -p api routes::oauth::tests --features test`
- `cargo test -p config cors_config`
- `git diff --check`
- `cargo check -p api --features test`

### DIFF
--- a/crates/api/src/routes/oauth.rs
+++ b/crates/api/src/routes/oauth.rs
@@ -26,8 +26,7 @@ use utoipa::ToSchema;
 
 const FRONTEND_CALLBACK_ALLOWED_ORIGINS_ENV: &str = "FRONTEND_CALLBACK_ALLOWED_ORIGINS";
 const OAUTH_CALLBACK_PATH: &str = "/auth/callback";
-const PRIVATE_CHAT_UNIVERSAL_LINK_HOST: &str = "app.privatechat.com";
-const MOBILE_FRONTEND_CALLBACK_SCHEMES: &[&str] = &["nearprivatechat", "privatechat"];
+const MOBILE_FRONTEND_CALLBACK_SCHEMES: &[&str] = &["nearprivatechat"];
 const MOBILE_FRONTEND_CALLBACK_HOST: &str = "auth";
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -137,7 +136,7 @@ fn validate_frontend_callback_url(
     let trimmed_url = raw_url.trim();
     let url = Url::parse(trimmed_url).map_err(|_| FrontendCallbackValidationError::Malformed)?;
 
-    if is_allowed_mobile_callback_url(&url)? || is_private_chat_universal_link_callback_url(&url)? {
+    if is_allowed_mobile_callback_url(&url)? {
         return Ok(normalize_validated_callback_url(url));
     }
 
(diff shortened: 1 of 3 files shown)


crates/api/src/routes/oauth.rs
```diff
@@ -235,6 +325,10 @@ fn url_origin(url: &Url) -> String {
     url.origin().ascii_serialization()
 }
 
+fn normalize_validated_callback_url(url: Url) -> String {
+    trim_trailing_slash(url.as_str())
```
walnut-the-cat: here, entire serialized URL, including query, becomes string to remove '/'. If URL contains state, e.g. state=abc/, it will trim / and can cause state mismatch. 

maybe the intention was to trip only url path.

crates/api/src/routes/oauth.rs
```diff
@@ -126,40 +130,65 @@ fn validate_frontend_callback_url(
     raw_url: &str,
     allowed_origins: Option<&[String]>,
 ) -> Result<String, FrontendCallbackValidationError> {
+    let trimmed_url = raw_url.trim();
+    let url = Url::parse(trimmed_url).map_err(|_| FrontendCallbackValidationError::Malformed)?;
+
+    if is_allowed_mobile_callback_url(&url)? || is_private_chat_universal_link_callback_url(&url)? {
```
walnut-the-cat: By doing this, we are bypassing allowed origin check below. Any bypass by default must be carefully considered and I am not sure how safe it is to make exception for app.privatechat.com when we don't own the domain. Basically we will be sending user session token and giving power for it to do further sensitive stuff.

crates/api/src/routes/oauth.rs
```diff
@@ -25,6 +25,9 @@ use url::{Host, Url};
 use utoipa::ToSchema;
 
 const FRONTEND_CALLBACK_ALLOWED_ORIGINS_ENV: &str = "FRONTEND_CALLBACK_ALLOWED_ORIGINS";
+const OAUTH_CALLBACK_PATH: &str = "/auth/callback";
+const MOBILE_FRONTEND_CALLBACK_SCHEMES: &[&str] = &["nearprivatechat"];
```
think-in-universe: Considering chat-api is also the backend of private agent hosting, we may use a more generic scheme in my opinion like `nearai`.

env.example
```diff
@@ -64,7 +64,8 @@ FRONTEND_URL=http://localhost:3000
 # Optional comma-separated exact bare origins allowed for user-provided OAuth frontend_callback.
 # If unset, callback origins are unrestricted; this is useful for staging frontend tests.
 # Entries must not include a path, query string, or fragment.
-# Remote callback origins must use HTTPS.
+# Remote callback origins must use HTTPS. The iOS/Android custom-scheme callback
+# is separately restricted to nearprivatechat://auth.
```
think-in-universe:
```suggestion
# is separately restricted to nearai://auth.
```

think-in-universe (changes requested):
Looks good except the scheme name convention

think-in-universe (approved):

PR #304, opened 2026-06-10
fix: SSE observability, TEE node placement, and backup_url bypass for migration

Longarithm:
## Summary

- **SSE stall detection**: replace the single 600s silent timeout with 30s stall-detection warnings visible in Datadog. Total budget enforced by reqwest `.timeout()`
- **Keepalive logging**: log SSE comment lines (`: keep-alive`) at INFO to fill the ~55s Datadog gap during backup/import
- **TEE node placement**: pass `node_policy: {version: 1, tee: "ANY_TEE"}` in the CrabShack import body so migrated instances land on TEE nodes
- **`backup_url` parameter**: optional field in request body — if provided, skip the entire backup SSE phase and use the URL directly. Sidesteps compose-api SSE stalls without needing `tokio::spawn` or `reuse_backup` complexity

## Context

During E2E migration testing (sessions 4-6), the backup SSE stream through chat-api stalls silently — likely because connection drops cancel the handler, which drops the reqwest stream to compose-api, leaving compose-api in a bad state. Direct compose-api `/backup` calls work fine.

The `backup_url` param lets the admin obtain the backup URL directly from compose-api, then pass it to `/migrate` to skip the problematic SSE relay entirely.

Companion PR: https://github.com/Near-One/ai-infra-agent-hosting/pull/327 (CrabShack side — accept `node_policy` in import, default to `ANY_TEE`)

## Test plan

- [x] `cargo check -p api` / `cargo fmt` / `cargo clippy` clean
- [x] 101 unit tests pass
- [ ] Deploy to staging, trigger migration with and without `backup_url`
- [ ] Verify stall-detect WARN and keepalive INFO lines in Datadog
- [ ] Verify migrated instance lands on TEE node (`staging-tdx-cvm-01`)

### DIFF
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -4036,13 +4036,13 @@ pub async fn admin_migrate_instance(
         }
     };
 
-    // Verify TEE placement — query CrabShack for the created instance and check tee_platform.
+    // Best-effort TEE placement check — read actual node's tee_platform, not the echoed policy.
     let instance_check_url = format!(
         "{}/instances/{}",
         crabshack_manager.url.trim_end_matches('/'),
         urlencoding::encode(&instance_name)
     );
-    if let Ok(resp) = app_state
+    match app_state
         .http_client
         .get(&instance_check_url)
         .bearer_auth(&crabshack_manager.token)
@@ -4050,18 +4050,33 @@ pub async fn admin_migrate_instance(
         .send()
         .await
     {
-        if let Ok(body) = resp.json::<serde_json::Value>().await {
-            let tee = body
-                .pointer("/node_policy/tee")
-                .and_then(|v| v.as_str())
-                .unwrap_or("unknown");
-            if tee == "NO_TEE" {
-                tracing::warn!(
-                    "Migrate: instance landed on non-TEE node, node_policy.tee={}, instance_id={}. \
-                     CrabShack may not support node_policy yet — check deploy ordering.",
-                    tee, id,
-                );
+        Ok(resp) if resp.status().is_success() => {
+            if let Ok(body) = resp.json::<serde_json::Value>().await {
+                let has_tee = body
+                    .pointer("/running_platform/tee_platform")
+                    .is_some_and(|v| !v.is_null());
+                if !has_tee {
+                    tracing::warn!(
+                        "Migrate: instance may not be on a TEE node (running_platform.tee_platform absent), \
+                         instance_id={}",
+                        id,
+                    );
+                }
             }
+        }
+        Ok(resp) => {
+            tracing::warn!(
+                "Migrate: TEE placement check failed: status={}, instance_id={}",
+                resp.status(),
+                id,
+            );
+        }
+        Err(e) => {
+            tracing::warn!(
+                "Migrate: TEE placement check failed: error={}, instance_id={}",
+                e,
+                id,
+            );
         }
     }
 
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -3708,7 +3708,7 @@ pub async fn admin_migrate_instance(
         .post(&backup_url)
         .bearer_auth(&manager.token)
         .json(&backup_body)
-        .timeout(std::time::Duration::from_secs(600))
+        .timeout(SSE_MIGRATION_TOTAL_TIMEOUT)
         .send()
         .await
         .map_err(|e| {
@@ -3883,7 +3883,7 @@ pub async fn admin_migrate_instance(
         "cpus": cpus,
         "storage_size": storage_size,
         "extra_env": extra_env,
-        "node_policy": { "tee": "ANY_TEE" },
+        "node_policy": { "version": 1, "tee": "ANY_TEE" },
     });
 
     let import_resp = app_state
@@ -3891,7 +3891,7 @@ pub async fn admin_migrate_instance(
         .post(&import_url)
         .bearer_auth(&crabshack_manager.token)
         .json(&import_body)
-        .timeout(std::time::Duration::from_secs(600))
+        .timeout(SSE_MIGRATION_TOTAL_TIMEOUT)
         .send()
         .await
         .map_err(|e| {
@@ -4019,6 +4019,7 @@ pub async fn admin_migrate_instance(
 /// Maximum buffer size for SSE parsing (1 MB).
 const SSE_BUFFER_LIMIT: usize = 1024 * 1024;
 /// Total time budget for an SSE migration stream (backup or import).
+/// Applied as reqwest's per-request .timeout() on the backup/import POST calls.
 const SSE_MIGRATION_TOTAL_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(600);
 /// How long to wait for a single chunk before logging a stall warning.
 /// Compose-api sends keepalives every 15s, so 30s without any data means trouble.
@@ -4049,16 +4050,9 @@ async fn parse_sse_for_backup_url(
             }
             Ok(None) => break,
             Err(_) => {
-                let silent_secs = last_chunk_at.elapsed().as_secs_f64();
-                if start.elapsed() >= SSE_MIGRATION_TOTAL_TIMEOUT {
-                    anyhow::bail!(
-                        "Timed out waiting for backup SSE ({}s total, no data for {:.0}s, {} chunks received), instance_id={}",
-                        SSE_MIGRATION_TOTAL_TIMEOUT.as_secs(), silent_secs, chunk_count, instance_id,
-                    );
-                }
                 tracing::warn!(
                     "Backup SSE: no data for {:.0}s (total elapsed={:.1}s, chunks={}), instance_id={}",
-                    silent_secs, start.elapsed().as_secs_f64(), chunk_count, instance_id,
+                    last_chunk_at.elapsed().as_secs_f64(), start.elapsed().as_secs_f64(), chunk_count, instance_id,
                 );
                 continue;
             }
@@ -4162,16 +4156,9 @@ async fn parse_sse_for_import_result(
             }
             Ok(None) => break,
             Err(_) => {
-                let silent_secs = last_chunk_at.elapsed().as_secs_f64();
-                if start.elapsed() >= SSE_MIGRATION_TOTAL_TIMEOUT {
-                    anyhow::bail!(
-                        "Timed out waiting for import SSE ({}s total, no data for {:.0}s, {} chunks received), instance_id={}",
-                        SSE_MIGRATION_TOTAL_TIMEOUT.as_secs(), silent_secs, chunk_count, instance_id,
-                    );
-                }
                 tracing::warn!(
                     "Import SSE: no data for {:.0}s (total elapsed={:.1}s, chunks={}), instance_id={}",
-                    silent_secs, start.elapsed().as_secs_f64(), chunk_count, instance_id,
+                    last_chunk_at.elapsed().as_secs_f64(), start.elapsed().as_secs_f64(), chunk_count, instance_id,
                 );
                 continue;
             }
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -3656,16 +3656,33 @@ pub async fn admin_migrate_instance(
     let image = if let Some(img) = image_override {
         img
     } else if has_backup_url {
+        let compose_image = compose_data
+            .get("image")
+            .and_then(|v| v.as_str())
+            .unwrap_or("");
         let default_image = services::agent::service::get_image_for_service_type(
             service_type,
             hosting_config.as_ref(),
         );
-        let mut fallback = default_image;
-        if fallback == "docker.io/nearaidev/ironclaw-dind:latest" {
-            fallback = "docker.io/nearaidev/ironclaw-dind:0.29.1".to_string();
-        }
+        let version_from_compose = compose_image
+            .rsplit_once(':')
+            .map(|(_, tag)| tag)
+            .filter(|t| !t.is_empty() && *t != "latest");
+        let fallback = if let Some(ver) = version_from_compose {
+            let base = default_image
+                .rsplit_once(':')
+                .map(|(b, _)| b)
+                .unwrap_or(&default_image);
+            format!("{}:{}", base, ver)
+        } else {
+            let mut img = default_image;
+            if img == "docker.io/nearaidev/ironclaw-dind:latest" {
+                img = "docker.io/nearaidev/ironclaw-dind:0.29.1".to_string();
+            }
+            img
+        };
         tracing::info!(
-            "Migrate: backup_url provided, using default image={}, instance_id={}",
+            "Migrate: backup_url provided, image derived from compose metadata={}, instance_id={}",
             fallback,
             id
         );
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -3550,6 +3550,12 @@ pub async fn admin_migrate_instance(
     // can reach the running container.
     let image_override = request.image.filter(|s| !s.is_empty());
     let provided_backup_url = request.backup_url.filter(|s| !s.is_empty());
+    let has_backup_url = provided_backup_url.is_some();
+    if let Some(ref url) = provided_backup_url {
+        if !url.starts_with("https://") {
+            return Err(ApiError::bad_request("backup_url must use https scheme"));
+        }
+    }
     let mem_limit = compose_data
         .get("mem_limit")
         .and_then(|v| v.as_str())
@@ -3569,6 +3575,13 @@ pub async fn admin_migrate_instance(
         .unwrap_or("unknown");
 
     let was_running = compose_status != "stopped" && compose_status != "exited";
+
+    if has_backup_url && was_running {
+        return Err(ApiError::bad_request(
+            "backup_url requires the legacy instance to be stopped first — \
+             stop it before taking an external backup to avoid data loss",
+        ));
+    }
 
     // Preflight: ironclaw instances must have SECRETS_MASTER_KEY available.
     // compose-api reads it via docker exec at startup — stopped containers are
@@ -3594,7 +3607,7 @@ pub async fn admin_migrate_instance(
     // When backup_url is provided, skip starting the instance and querying /version —
     // the admin already obtained the backup externally (e.g. direct compose-api call)
     // and the instance should stay stopped to avoid accepting writes after the backup.
-    if provided_backup_url.is_none() && !was_running {
+    if !has_backup_url && !was_running {
         tracing::info!(
             "Migrate: instance stopped, starting before backup, elapsed={:.1}s, instance_id={}",
             migrate_start.elapsed().as_secs_f64(),
@@ -3642,7 +3655,7 @@ pub async fn admin_migrate_instance(
     // otherwise fall back to the configured default.
     let image = if let Some(img) = image_override {
         img
-    } else if provided_backup_url.is_some() {
+    } else if has_backup_url {
         let default_image = services::agent::service::get_image_for_service_type(
             service_type,
             hosting_config.as_ref(),
@@ -3806,52 +3819,55 @@ pub async fn admin_migrate_instance(
         }
     };
 
-    tracing::info!(
-        "Migrate: backup complete, stopping legacy instance, elapsed={:.1}s, instance_id={}",
-        migrate_start.elapsed().as_secs_f64(),
-        id,
-    );
-    let stop_url = format!("{}/instances/{}/stop", compose_api_url, encoded_name);
-    let stop_resp = app_state
-        .http_client
-        .post(&stop_url)
-        .bearer_auth(&manager.token)
-        .timeout(std::time::Duration::from_secs(60))
-        .send()
-        .await;
-
-    let stop_failed = match &stop_resp {
-        Err(e) => {
-            tracing::error!(
-                "Failed to stop instance on compose-api: instance_id={}, error={}",
-                id,
-                e
-            );
-            true
-        }
-        Ok(resp) if !resp.status().is_success() => {
-            tracing::error!(
-                "Compose-api stop returned non-success: instance_id={}, status={}",
-                id,
-                resp.status()
-            );
-            true
-        }
-        _ => false,
-    };
-    if stop_failed {
-        restore_on_failure(
-            &app_state,
-            compose_api_url,
-            &encoded_name,
-            &manager.token,
-            was_running,
-            "stop",
-            &instance_name,
+    // Skip stop when backup_url was provided — instance is already stopped (enforced above).
+    if !has_backup_url {
+        tracing::info!(
+            "Migrate: backup complete, stopping legacy instance, elapsed={:.1}s, instance_id={}",
+            migrate_start.elapsed().as_secs_f64(),
+            id,
         );
-        return Err(ApiError::internal_server_error(
-            "Failed to stop legacy instance before import",
-        ));
+        let stop_url = format!("{}/instances/{}/stop", compose_api_url, encoded_name);
+        let stop_resp = app_state
+            .http_client
+            .post(&stop_url)
+            .bearer_auth(&manager.token)
+            .timeout(std::time::Duration::from_secs(60))
+            .send()
+            .await;
+
+        let stop_failed = match &stop_resp {
+            Err(e) => {
+                tracing::error!(
+                    "Failed to stop instance on compose-api: instance_id={}, error={}",
+                    id,
+                    e
+                );
+                true
+            }
+            Ok(resp) if !resp.status().is_success() => {
+                tracing::error!(
+                    "Compose-api stop returned non-success: instance_id={}, status={}",
+                    id,
+                    resp.status()
+                );
+                true
+            }
+            _ => false,
+        };
+        if stop_failed {
+            restore_on_failure(
+                &app_state,
+                compose_api_url,
+                &encoded_name,
+                &manager.token,
+                was_running,
+                "stop",
+                &instance_name,
+            );
+            return Err(ApiError::internal_server_error(
+                "Failed to stop legacy instance before import",
+            ));
+        }
     }
 
     tracing::info!(
@@ -3904,6 +3920,8 @@ pub async fn admin_migrate_instance(
             ApiError::internal_server_error("No CrabShack manager configured")
         })?;
 
+    // Deploy ordering: node_policy requires CrabShack PR #327 to be deployed first.
+    // Without it, CrabShack ignores the field (pre-#327 import silently drops unknown fields).
     let import_url = format!("{}/instances/import", crabshack_manager.url);
     let import_body = serde_json::json!({
         "name": instance.name,
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -3341,6 +3341,9 @@ pub struct MigrateInstanceRequest {
     /// Skip the backup phase entirely and use this presigned URL instead.
     /// Useful when backup was obtained directly from compose-api (e.g. the
     /// SSE stream stalls through chat-api but works when called directly).
+    /// The backup must be age-encrypted with the recipient derived from the
+    /// user's stored passphrase + instance name. The legacy instance will NOT
+    /// be started — stop it before taking the external backup to avoid data loss.
     pub backup_url: Option<String>,
 }
 
@@ -3588,7 +3591,10 @@ pub async fn admin_migrate_instance(
         ));
     }
 
-    if !was_running {
+    // When backup_url is provided, skip starting the instance and querying /version —
+    // the admin already obtained the backup externally (e.g. direct compose-api call)
+    // and the instance should stay stopped to avoid accepting writes after the backup.
+    if provided_backup_url.is_none() && !was_running {
         tracing::info!(
             "Migrate: instance stopped, starting before backup, elapsed={:.1}s, instance_id={}",
             migrate_start.elapsed().as_secs_f64(),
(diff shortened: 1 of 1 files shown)


PierreLeGuen (changes requested):
This moves the legacy→CrabShack migration into a `tokio::spawn` background task returning 202, replaces the 600s SSE chunk timeout with 30s stall detection, and adds an opt-in `reuse_backup` flag. The refactor is mechanically sound — the age keypair is deterministically derived from `(backup_passphrase, instance_name)` so reused backups stay decryptable, and failure paths consistently call `restore_on_failure` — but the new `provisioning` lifecycle introduces operational hazards that should be addressed before merge.

**Blocking**

- `crates/api/src/routes/admin.rs:3616` / `3674` — An instance stuck in `provisioning` has no recovery path. The endpoint persists `provisioning` and then relies on an untracked `tokio::spawn` to finish or reset it. If the process restarts, deploys, or the task panics after the 202 is returned — or the failure-path `update_instance_status` at 3676 fails (it's only logged) — the status stays `provisioning` forever. The guard at `admin.rs:3381` then permanently rejects re-migration, and `crates/api/src/routes/agents.rs:1528` rejects all of the instance's agent API keys (`status != "active"`), bricking the instance until manual DB surgery. Add a startup reconciliation sweep (reset instances in `provisioning` with `change_reason='migration_started'`) and/or treat `provisioning` older than the migration time budget as stale, plus a small retry loop on the final status transitions.
- `crates/api/src/routes/admin.rs:3381` vs `3618` — The `provisioning` guard and the status write are a non-atomic check-then-set separated by several awaits, including a compose-api round trip with a 30s timeout. Two concurrent POSTs to `/migrate` both pass the guard and spawn duplicate full migrations (duplicate backup, double stop/import, conflicting status updates). Gate atomically, e.g. `UPDATE agent_instances SET status='provisioning' WHERE id=$1 AND status <> 'provisioning'` and return a conflict when 0 rows are affected.
- `crates/api/src/routes/admin.rs:4199-4264` (`fetch_latest_age_backup_url`) — `reuse_backup` accepts the latest age-encrypted backup with no recency bound, despite the doc comment and bail message saying "recent". An admin retrying weeks after a failed attempt silently imports a stale backup, discarding everything the user wrote since (the instance keeps serving and accumulating data between attempts). Enforce a max age (e.g. the 7200s presigned expiry used at creation) or require an explicit backup id in the request.
- `crates/api/src/routes/admin.rs:3618` + `crates/api/src/routes/agents.rs:1528` — Setting `provisioning` for the entire migration is a user-visible behavior change: agent API key verification fails the moment migration begins, even though the legacy instance keeps serving throughout the long backup phase (migration previously never touched status). If freezing the instance for the whole window is intended, document it; otherwise only set `provisioning` around the stop→import→DB-update window.

**Non-blocking**

- `crates/api/src/routes/admin.rs:4288-4292` — `SSE_MIGRATION_TOTAL_TIMEOUT` is documented as a total time budget but is only checked when a ≥30s stall occurs; a stream that keeps emitting keepalives is bounded only by the reqwest `.timeout(600)`, which surfaces as a generic stream error instead of the descriptive bail. Check `start.elapsed()` every iteration, or document that the request timeout is the real bound.
- `crates/api/src/routes/admin.rs:4269-4283` (`is_age_key_encrypted`) — If the presigned storage ignores the `Range` header and returns 200 with the full body, the function falls through and `resp.bytes()` buffers the entire backup (potentially GBs) into memory. Read only the first chunk via `bytes_stream()`, and accept either 200 or 206.
- No test coverage: the response contract changed (200 `status:"success"` → 202 `status:"started"`, new `provisioning` rejection, new `reuse_backup` field) and nothing in `crates/api/tests/` exercises it. At minimum the preflight behavior (guard + 202 shape) is testable without a compose-api mock.

**Checks run:** `cargo check -p api`, `cargo fmt --all -- --check`, and `cargo clippy -p api --all-targets --all-features -- -D warnings` all passed; `cargo test -p api --lib --bins --features test` passed (101 tests). The integration suite wasn't run locally (requires a local Postgres); CI's Test Suite check is green on the PR head.

PierreLeGuen (changes requested):
This PR's observability improvements (stall WARNs, keepalive INFO logging) are sound and consistent with the logging rules — only IDs, counters, and elapsed times are logged. However, the total-timeout logic doesn't behave as described, and the `node_policy` payload looks malformed against the companion validator. Requesting changes for those two.

**Blocking / important**

- `crates/api/src/routes/admin.rs:4021-4025` (with the pre-existing `.timeout(600s)` at 3711 and 3894) — `SSE_MIGRATION_TOTAL_TIMEOUT` is only enforced inside the stall branch, but reqwest 0.12 applies the per-request timeout to the entire streamed body. Two consequences: (a) a healthy migration emitting keepalives still hard-fails at 600s, surfaced as a generic reqwest error through `chunk?`, so the loop's permissive design is dead letter; (b) the new diagnostic bail ("no data for Xs, N chunks received") is effectively unreachable — any stall crossing the 600s mark resolves `stream.next()` with reqwest's deadline error before the in-loop check fires. Decide which mechanism owns the total budget: remove or raise the request-level `.timeout()` so the in-loop logic (moved to fire unconditionally, not just on stall) is authoritative and its diagnostics actually reach Datadog, or drop the in-loop total check and document that reqwest enforces the cap.
- `crates/api/src/routes/admin.rs:3886` — the import payload sends `node_policy: {"tee": "ANY_TEE"}`, but the companion CrabShack validator (Near-One/ai-infra-agent-hosting#327) requires `node_policy.version` to be an integer, so migration imports would fail validation instead of landing on a TEE node. Send `{"version": 1, "tee": "ANY_TEE"}`, or omit `node_policy` and rely on the companion's `ANY_TEE` default. Separately, this creates a deploy-ordering dependency: if the currently deployed CrabShack rejects unknown JSON fields, every migration 400s until the companion ships (recoverable via `restore_on_failure`, but still a migration outage) — please confirm tolerance and sequence the deploys. A config escape hatch for environments without TEE capacity would also be safer than a hardcoded value.

**Non-blocking**

- `crates/api/src/routes/admin.rs:4044-4088` and `4157-4201` — the stall-detection loop, total-timeout bail, and keepalive logging are duplicated verbatim between `parse_sse_for_backup_url` and `parse_sse_for_import_result` (differing only in the "Backup"/"Import" label). Extract a shared helper so the two timeout state machines don't drift.
- `crates/api/src/routes/admin.rs:4045-4047` — `last_chunk_at` is updated before `chunk?` propagates a transport error, so a failing read counts as "data received". Harmless in practice since the error bails, but reorder for consistency.
- No test coverage for the new stall/total-timeout state machine. The functions take `reqwest::Response` directly, which makes them untestable; accepting `impl Stream<Item = reqwest::Result<Bytes>>` would allow unit tests for stall-warn, total-timeout bail, keepalive handling, and complete/error events.

**Checks run locally**

- `cargo fmt --all -- --check` — clean
- `cargo check -p api --features test` — passed
- `cargo clippy -p api --all-targets --features test -- -D warnings` — passed
- `cargo test -p api --lib --features test` — passed (101 tests); integration tests requiring PostgreSQL were skipped, and nothing in the suite exercises the migration SSE paths touched here
- `gh pr view/diff/checks` — GitHub Test Suite was still pending at review time; the companion CrabShack repo was not inspectable from this environment, so the deployed import endpoint's unknown-field tolerance is unverified

PierreLeGuen (changes requested):
This PR is well-constructed overall: the stall-detection refactor is sound (reqwest 0.12's per-request `.timeout()` bounds the entire body read, so the 600s total budget is preserved while chunk stalls become WARN logs), the keepalive comment-line handling is correct, new log lines contain only IDs/durations/counts per the privacy rules, and `node_policy.version: 1` is correctly a JSON integer. One issue should be fixed before merge; the rest are non-blocking.

**Blocking**

- `crates/api/src/routes/admin.rs:3591-3624` — when `backup_url` is provided, the start-if-stopped block still boots a stopped legacy instance before the `provided_backup_url` branch at line 3698, even though the instance is only needed for the backup phase and the `/version` query. If an admin stopped the legacy instance precisely to freeze writes before taking an external backup, `/migrate` restarts it, lets it accept writes, and then imports the older backup — silently discarding those writes. Move the `backup_url` decision ahead of the start block and skip the start when a backup is supplied (require an image override or use the existing fallback image path instead of querying `/version`); consider requiring the legacy instance to be stopped before importing an external backup.

**Non-blocking**

- `crates/api/src/routes/admin.rs:4043-4247` — two related robustness points in the SSE parse loops. (1) They no longer have an internal termination bound: the only exit on a stalled stream is the caller's reqwest `.timeout()` (set at 3724/3908); a future caller that forgets it gets an infinite WARN-every-30s loop. (2) When the 600s budget does expire, the error surfaces via `chunk?` as a generic reqwest timeout instead of the old descriptive "Timed out waiting for SSE chunk (600s, received N chunks in Xs)" message. Both are fixed by one change: bail inside the loop when `start.elapsed() > SSE_MIGRATION_TOTAL_TIMEOUT` with the chunk count and elapsed time in the message.
- `crates/api/src/routes/admin.rs:4043-4247` — the stall/keepalive loop body is duplicated verbatim between `parse_sse_for_backup_url` and `parse_sse_for_import_result` (only the "Backup"/"Import" label differs). Extract a shared helper so the two timeout state machines can't drift — and that helper, taking an `impl Stream<Item = Result<Bytes>>`, would also make the keepalive, stall-resume, error-event, and complete-event paths unit-testable. The migrate endpoint currently has zero automated coverage, and this PR adds branching to it.
- `crates/api/src/routes/admin.rs:3900` — `node_policy: {version: 1, tee: "ANY_TEE"}` is hardcoded with no config gate, which creates a deploy-ordering coupling with the companion CrabShack PR #327: deployed first, the field is either silently ignored or rejected as an unknown field; once #327 ships (which defaults to ANY_TEE server-side) it's redundant. Note the deploy order in the runbook and consider making the policy configurable (e.g. `CRABSHACK_DEFAULT_NODE_POLICY`), especially for environments without TEE capacity.
- `crates/api/src/routes/admin.rs:3549,3698` — `backup_url` is accepted with no validation, not even an https-scheme check, and is forwarded for CrabShack to fetch (SSRF-by-proxy, mitigated by admin-only access). Misuse fails safely — the age recipient is HKDF-bound to `(passphrase, instance_name)` (`age_derivation.rs:114`), so a wrong or cross-instance backup fails decryption and `restore_on_failure` kicks in — but only after the legacy instance was stopped. Require `https://` and optionally pin the host to the expected storage endpoint before the destructive phase.
- `crates/api/src/routes/admin.rs:3341-3344` — the `backup_url` doc comment omits the two contract requirements an admin must meet: the backup must be encrypted to the age recipient derived from this user's stored passphrase + instance name, and writes between external backup creation and the migrate call are lost. Document both so the failure modes are discoverable from the OpenAPI schema.

**Local checks**: `cargo fmt --check`, `cargo check -p api`, and `cargo clippy -p api --all-targets -- -D warnings` all clean; `cargo test -p api --lib` passed (101 tests). Integration tests were skipped (no local PostgreSQL); the GitHub "Test Suite" check was still pending. Verified the checkout matches the PR head (`925f5f6`) and cross-checked the `node_policy` import contract against CrabShack PR #327.

PierreLeGuen (changes requested):
The SSE observability changes look sound — I verified in the reqwest 0.12.24 source that the per-request `.timeout()` is enforced on streamed body reads, so removing the in-loop hard timeout doesn't make the stall-detect loop unbounded, and the keepalive/stall log lines contain only IDs, counters, and durations. The substantive issues are concentrated in the new `backup_url` bypass path, which is a destructive (stop-legacy-instance) flow with no test coverage.

**Blocking / important**

1. **Unconditional `/stop` breaks the documented `backup_url` workflow** (`crates/api/src/routes/admin.rs:3814-3855`). The intended flow has the instance already stopped before the external backup is taken, yet the handler still POSTs `/instances/{name}/stop` and hard-fails the migration on any non-2xx ("Failed to stop legacy instance before import"). If compose-api isn't idempotent for stop-on-stopped, every `backup_url` migration — the headline feature of this PR — fails here. Skip the stop (or tolerate its failure) when `provided_backup_url.is_some() && !was_running`, and add a mocked test that this path calls neither `/start`, `/backup`, nor `/stop`.

2. **No guard against a still-running instance with `backup_url`** (`admin.rs:3552`, doc at `admin.rs:3344-3346`). The doc says "stop it before taking the external backup to avoid data loss," but nothing enforces it: if the instance is still running, the handler stops it at 3814 and imports the externally-taken (now stale) backup, silently discarding every write made since that backup. Return `bad_request` when `provided_backup_url.is_some() && was_running` (or at minimum emit a prominent WARN and document the accepted staleness). The doc comment should also clarify that "will NOT be started" only applies when the instance was already stopped.

3. **`backup_url` is accepted with zero validation** (`admin.rs:3552`/`3720`). Any non-empty string is forwarded as-is into CrabShack's backup download — an admin-side SSRF vector into CrabShack's network, and a typo only surfaces as a cryptic import-side failure *after* the legacy instance has been stopped. Parse with `url::Url` (already a dependency), require an `https` scheme, and fail before any destructive action.

4. **Hardcoded `node_policy: {version: 1, tee: "ANY_TEE"}` creates a deploy-ordering dependency** (`admin.rs:3922`). It's sent unconditionally on every import, and CrabShack's import validator is strict — if this ships before Near-One/ai-infra-agent-hosting#327, all migrations (not just `backup_url` ones) fail at import. There's also no escape hatch for non-TEE placement. Call out the ordering in the deploy notes, and consider making the policy configurable.

**Minor**

- `admin.rs:3645-3659` — the `backup_url` branch skips the `/version` query even when the instance is running, so the migrated instance comes up on the pinned default (`ironclaw-dind:0.29.1`), which may be older than what the legacy instance ran. Still attempt the best-effort version query when `was_running`, or document that `image` should be passed alongside `backup_url`.
- `admin.rs:3651-3653` / `3707-3710` — the `docker.io/nearaidev/ironclaw-dind:latest` → `:0.29.1` pin is duplicated verbatim in two branches; extract a helper/constant so the next bump doesn't miss one site.
- `admin.rs:4088-4094` / `4194-4200` — when the total timeout does fire, the descriptive stall diagnostic ("no data for Ns, received N chunks") is only ever a WARN; the caller sees a generic reqwest transport error. Consider an in-loop `elapsed > SSE_MIGRATION_TOTAL_TIMEOUT` bail that preserves the chunk count and elapsed time in the returned error.
- `admin.rs:4110-4117` / `4216-4223` — keepalive comment lines log at INFO roughly every 15s per migration; consider `debug!` for keepalives and reserve INFO for `stage` events.
- No tests cover the `backup_url` branching, keepalive handling, or stall-detect paths, and `parse_sse_for_backup_url`/`parse_sse_for_import_result` take `reqwest::Response` directly, making them untestable. Their loop bodies are near-identical — extracting a shared helper over a generic byte stream would dedupe the code and make all of these paths unit-testable in one move.

**Checks run locally:** `cargo fmt --check` clean; `cargo check -p api` passes; `cargo clippy -p api --all-targets --all-features -- -D warnings` clean; `cargo test -p api --lib --features test` — 101 tests passed (integration tests requiring PostgreSQL not run). Also verified in vendored reqwest 0.12.24 source that the per-request timeout wraps the response body stream. GitHub "Test Suite" CI was still in progress at review time.

PierreLeGuen (changes requested):
This is a focused, admin-only change to the migration endpoint, and the core mechanics check out: with `backup_url`, `was_running` is forced false so the `restore_on_failure(..., "start", ...)` path is correctly a no-op, the per-instance age-encryption recipient means a cross-wired backup URL fails to decrypt rather than restoring the wrong customer's data, and all new log lines carry only IDs, timings, counts, and image names (the presigned URL is never logged). The SSE timeout change is also sound: the old 600s per-chunk timeout was effectively dead code since the reqwest per-request `.timeout(600s)` (present on `main` at the backup and import POSTs) already capped totals through body completion, so swapping it for a 30s non-fatal stall warning improves observability without changing the worst-case budget, and the warn loop can't spin forever because the reqwest deadline surfaces as a stream error via `chunk?`.

Two issues should be addressed before merge:

**Important**

- `crates/api/src/routes/admin.rs:3658-3672` — In the `backup_url` path the image is resolved from the configured default (or the hardcoded pin) because `/version` can't be queried on a stopped instance. The compose metadata already contains the legacy `image` used earlier to infer service type, so the version tag can be derived from `compose_data["image"]` instead. As written, a stopped-instance migration can silently become an unplanned runtime upgrade, or import an older backup into an incompatible image, with nothing prompting the admin to pass an `image` override. Derive the tag from the compose metadata, or require an explicit `image` override when the version can't be inferred (at minimum, WARN that the version was not verified).
- `crates/api/src/routes/admin.rs:3923-3941` — `node_policy: {version: 1, tee: "ANY_TEE"}` is sent best-effort and, per the adjacent comment, silently ignored by pre-#327 CrabShack. The code still treats the import as successful and updates the DB, so if deploy ordering slips, migrations report success while landing on non-TEE nodes — and TEE placement is the security property this change exists to guarantee. Verify placement from the import result (or a post-import query) and fail or WARN when the node is not a TEE node, or gate on a CrabShack capability/version, rather than relying on a code comment about deploy ordering.

**Non-blocking**

- `crates/api/src/routes/admin.rs:3554-3558` — `backup_url` validation is only `starts_with("https://")`; the string is forwarded verbatim to CrabShack, which downloads it (an admin-supplied SSRF vector, and malformed URLs fail late at import). Parse with the already-imported `url` crate, require an https scheme and non-empty host, and consider an allowlist of expected storage hosts.
- `crates/api/src/routes/admin.rs:3663-3666`, `3720-3723` — the `docker.io/nearaidev/ironclaw-dind:latest` → `0.29.1` pin is now duplicated in two branches; extract it to a constant (or fold into `get_image_for_service_type`) so the next version bump doesn't miss one site.
- The two SSE parse functions duplicate their loop bodies nearly verbatim, differing only in labels and final data extraction. Extracting a shared helper over `impl Stream<Item = Result<Bytes>>` would keep stall detection and keepalive handling in sync between the paths and make them unit-testable. Relatedly, when the reqwest 600s deadline fires it surfaces as a generic transport error — including elapsed time and chunk count in the error context would preserve the diagnostics the old hard-timeout message carried.
- `crates/api/src/routes/admin.rs:3579-3584` — the "must be stopped" guard is a snapshot check; the instance can be started between the status fetch and the import. Since stop is idempotent, issuing it unconditionally (instead of skipping when `backup_url` is set) would close the race cheaply.
- `crates/api/src/routes/admin.rs:3594-3605` — the ironclaw SECRETS_MASTER_KEY preflight error tells the admin to "start all instances and restart compose-api before migration", which contradicts the `backup_url` precondition that the instance be stopped. Branch the message when `backup_url` is set to describe the correct ordering.
- The new validation branches (non-https rejection, running-instance rejection) are pure and testable but untested; the migrate endpoint has no test coverage at all (pre-existing gap this PR extends). Adding cases for the two new 400 paths would be cheap.

Checks run locally on PR head `d6190e9`: `cargo fmt --check` (clean), `cargo check -p api` (clean), `cargo clippy -p api --all-targets -- -D warnings` (clean), `cargo test -p api --lib` (101 passed), `cargo test -p services image --lib` (21 passed). GitHub CI "Test Suite" is green on the head commit. Integration/E2E migration tests were not run locally (they require live PostgreSQL and compose-api credentials) and are deferred to CI and the staging test plan.

PierreLeGuen (changes requested):
This iteration addresses the image-derivation feedback from the previous round, and the SSE observability changes are sound — the shared `http_client` has no global timeout, so the per-request 600s deadline bounds the stream and the 30s stall-warn loop can't spin forever, and the new log lines carry only IDs/counters/durations/image names. Two issues from the `backup_url` and TEE-placement changes still need to be fixed before merge.

**Important**

- `crates/api/src/routes/admin.rs:3940-3957` — `node_policy: {version: 1, tee: "ANY_TEE"}` is sent unconditionally, and per the comment at 3940 a pre-#327 CrabShack silently drops unknown fields. The handler then reports success and updates the DB regardless of where the instance landed. TEE placement is the security property in this PR's title, yet nothing verifies it: if deploy ordering slips, migrations succeed onto non-TEE nodes. This was flagged last round and the response is a comment, not enforcement. Fail closed — verify placement from the import SSE result or a post-import instance query (or gate on a CrabShack capability/version check) before updating the DB.

- `crates/api/src/routes/admin.rs:3415-3448` (with 3552) — when the user has no stored passkey credentials, the handler generates and stores a fresh `backup_passphrase` before the request body is inspected. That was fine when this endpoint also created the backup, but with `backup_url` the backup already exists and cannot have been encrypted to the new recipient (the doc comment at 3344-3346 requires the *stored* passphrase), so decryption is guaranteed to fail late at CrabShack import — after up to 600s, with the legacy instance left stopped — and even rejected `backup_url` requests can mutate credential state first. Parse/validate the request before touching credentials, and when `backup_url` is set with no existing credentials, return `bad_request` up front (or add a preflight flow that issues the age recipient before the external backup is taken). Tests, please.

**Minor**

- `admin.rs:3554-3558` — `backup_url` validation is still just `starts_with("https://")` and the string is forwarded verbatim to CrabShack's downloader (admin-supplied SSRF vector, late failures on malformed URLs). Parse with the already-depended-on `url` crate, require https + non-empty host, and consider an allowlist of expected storage hosts.
- `admin.rs:3579-3584` / `3840` — the "must be stopped" guard is a one-time snapshot; if the instance is started between the status check and the import, the stop is skipped and you end up with the legacy instance running alongside the new one while the DB points at the new one, with a stale backup imported. compose-api stop is idempotent — issue it unconditionally instead of skipping it when `backup_url` is set.
- `admin.rs:3658-3689` — the tag-grafting heuristic assumes legacy compose tags exist in the CrabShack image repo, but the repos are named differently (that's why service-type inference at 3537-3540 exists), so a non-matching tag yields a nonexistent image and a late pull failure. Also, the INFO line at 3684 claims "image derived from compose metadata" even on the fallback branch. WARN when grafting a tag onto a different base repo and split the log message.
- `admin.rs:3678-3682` and `3737-3740` — the `ironclaw-dind:latest` → `0.29.1` pin is duplicated verbatim; the next version bump will likely miss one. Extract a constant or fold it into `get_image_for_service_type`.
- `admin.rs:3594-3605` — the SECRETS_MASTER_KEY preflight error tells the admin to "start all instances and restart compose-api before migration", which contradicts the `backup_url` precondition that the instance be stopped. Branch the message when `backup_url` is set to describe the correct ordering.
- No test references the migrate endpoint. The two new 400 guards (non-https `backup_url`, running-instance rejection) and the image derivation are cheaply unit-testable, and `parse_sse_for_backup_url`/`parse_sse_for_import_result` take `reqwest::Response` directly with near-identical loop bodies, keeping the keepalive/stall paths untestable. Add the 400-path tests and extract the derivation (ideally a shared SSE-parse helper over `impl Stream`) into testable functions.

**Checks run locally:** `cargo fmt --check`, `cargo check -p api`, and `cargo clippy -p api` all clean; `cargo test -p api --lib --features test` passed 101/101, matching the PR's test plan. GitHub CI "Test Suite" was still pending on the head commit at review time. Integration/E2E migration tests were skipped (require live PostgreSQL and compose-api/CrabShack credentials); the staging items in the test plan (TEE placement, Datadog log lines) remain unverified.

PierreLeGuen (changes requested):
This fix is well-constructed in most respects: the new log lines carry only IDs/durations/counts (no presigned URL leakage), the stall-detection rework is sound — reqwest 0.12's per-request `.timeout()` is a total deadline that covers body streaming, so the warn-and-continue loop cannot run past the 600s budget (the old per-chunk bail was effectively dead code) — and the `backup_url` path correctly refuses running instances, skips start/stop, and never restarts the legacy instance on failure (`was_running` is guaranteed false there, making every `restore_on_failure` call a no-op). Two issues need to be addressed before merge.

**Important**

- `crates/api/src/routes/admin.rs:4039-4066` — The post-import TEE placement verification is a no-op in exactly the scenario it was added to catch. It reads `/node_policy/tee` from the GET response — the *requested* policy echoed back, not actual placement — and warns only on the literal value `"NO_TEE"`. In the documented failure mode (pre-#327 CrabShack silently dropping `node_policy`), the field is absent, the value defaults to `"unknown"`, and no warning fires. The response status is never checked, and both `.send()` and `.json()` errors are swallowed by `if let Ok`, so an unreachable CrabShack or error body also passes silently. Fix: check `resp.status().is_success()`, read the actual placement field (`tee_platform` / `running_platform`) rather than the echoed policy, and warn (or fail, per the intended guarantee) on anything other than a confirmed TEE placement — including missing/unknown values and query failures.

- `crates/api/src/routes/admin.rs:3415-3428, 3562-3594, 3668-3699` — No tests cover any of the new branching logic, despite this endpoint being able to lose customer data if its guards regress. The repo has wiremock and an established admin-route test suite (`crates/api/tests/admin_tests.rs`). At minimum, cover: the three new 400 validations (backup_url without stored passkey credentials, backup_url with a running instance, non-HTTPS backup_url), that no `/start`/`/backup`/`/stop` calls are issued when `backup_url` is set, that the import body carries the provided URL/passphrase/image/`node_policy`, and the compose-metadata image derivation (which is pure and easily extracted into a testable function).

**Minor**

- `admin.rs:3677-3686` — Image derivation grafts the legacy compose image's tag onto the default image base, assuming that tag exists in the new repo, and doesn't filter tags containing `/` (a `registry:port/name` ref without a tag yields "tag" = `5000/name`, an invalid image ref). It fails safe — the import fails, the instance stays stopped, and the admin can retry with the explicit `image` override — but a one-line guard rejecting tags containing `/` (and a doc note that cross-repo tags must exist) is cheap insurance.
- `admin.rs:3688-3691` and `3747-3750` — The `ironclaw-dind:latest` → `0.29.1` pin is now duplicated in two branches; extract a shared constant so the next bump can't miss one site.
- `admin.rs:3564-3568` — `backup_url` validation is just `starts_with("https://")` before being forwarded verbatim to CrabShack, which fetches it — an admin-controlled SSRF surface into the hosting network. Low risk given admin-only access, but parse with `url::Url` (require https scheme + non-empty host) and consider an allowlist of expected storage hosts.
- `admin.rs:3582-3594` — The "instance must be stopped" guard uses a `compose_status` snapshot from earlier in the request, so an instance started concurrently would still be migrated from a stale backup. Acceptable for an admin-driven flow, but re-checking status immediately before import would make the data-loss guard real rather than advisory.
- `admin.rs:3604-3653` — When `backup_url` is set, the SECRETS_MASTER_KEY preflight error still says "start all instances and restart compose-api before migration," which is the wrong guidance for the external-backup flow. Branch the message.
- `admin.rs:4139-4343` — `parse_sse_for_backup_url` and `parse_sse_for_import_result` are near-verbatim duplicates and have already drifted (only the import parser checks the `event` field). Extract a shared helper parameterized by label, taking the response/stream so the keepalive/stall/complete/error paths become unit-testable.

**Checks run locally**
- `cargo fmt --check` — clean
- `cargo check -p api` — passed
- `cargo clippy -p api --all-targets -- -D warnings` — clean
- `cargo test -p api --lib` — 101 passed (no tests exercise the migrate endpoint or SSE parsers; full integration suite requires PostgreSQL)
- `git diff --check` — clean
- Verified reqwest 0.12.24 in Cargo.lock to confirm the per-request timeout bounds streaming-body reads
- GitHub "Test Suite" CI check was still pending at review time

PierreLeGuen (dismissed):
This PR touches only `crates/api/src/routes/admin.rs` (+283/−126): SSE stall warnings replacing the 600s silent per-chunk timeout, keepalive logging, `node_policy` TEE placement on import, and a `backup_url` bypass of the backup phase. The `backup_url` flow is well-guarded overall — rejected without stored passkey credentials, rejected unless the instance is already stopped, and the start/stop/restore steps correctly no-op (`was_running` is necessarily false). New log lines carry only IDs, counters, durations, and image names; the presigned URL is never logged. A few issues need addressing before merge.

**Blocking / important**

- `admin.rs:4039-4081` — The post-import TEE placement check is broken and silent. It reads `/running_platform/tee_platform`, but CrabShack's instance response exposes `running_platform.version` with no nested `tee_platform` field, so the check warns even for correctly placed TEE imports — the deploy-ordering safety net is non-functional as written. It should check `running_platform.version != null`. Independently, the check only emits `tracing::warn!` and the migration still returns `status: "success"`: a migration that lands customer data on a non-TEE node looks fully successful to the admin unless they're watching Datadog. At minimum surface the placement result in `MigrateInstanceResponse` (message or field); also note a JSON parse failure silently skips the check (`if let Ok(body)` has no else arm), and querying before the instance finishes starting can spuriously warn.
- `admin.rs:3564-3568` — `backup_url` validation is `starts_with("https://")` only, and the string is forwarded verbatim for CrabShack to fetch (admin-controlled SSRF into CrabShack's network; admin-only, so low risk, but unnecessary exposure). Worse, malformed URLs fail late at import, *after* the legacy instance is stopped. Parse with the already-depended-on `url` crate (https scheme + non-empty host), fail before any destructive action, and consider allowlisting the expected storage host.
- `admin.rs:3668-3699` — The image derivation grafts the legacy compose image's tag onto the *default* image base for the service type. If the legacy repo lineage differs from the configured default (e.g. an `ironclaw` → `ironclaw-dind` rename), `<default-base>:<legacy-tag>` may not exist, failing the import with the instance left stopped. Also, `rsplit_once(':')` on a tagless image with a registry port (`registry.example.com:5000/foo`) yields a bogus "tag" of `5000/foo` that passes the `!= "latest"` filter and produces an invalid reference. Reject tags containing `/`, warn when compose base ≠ default base, and fix the INFO log at `admin.rs:3695`, which claims "derived from compose metadata" even on the hardcoded `latest → 0.29.1` fallback branch.
- Tests — the PR adds four new request-contract rejection rules (missing credentials, non-https URL, running instance, empty-string normalization) and a non-trivial image-derivation heuristic to a destructive endpoint with zero existing test coverage; the test plan is entirely manual. The image derivation is pure logic and cheap to unit-test if extracted into a free function; the validation paths are testable with the existing wiremock-based admin harness. At minimum, cover the three new 400 paths and assert `/start`/`/backup`/`/stop` are not called when `backup_url` is set.

**Minor**

- `admin.rs:3688-3691` vs `3747-3750` — the `ironclaw-dind:latest → 0.29.1` pin is now duplicated verbatim in two branches; extract a shared const so the next version bump can't miss one.
- `admin.rs:3589-3594` — the stopped-instance guard is TOCTOU: status is read once well before import, so the instance can be started in between, reintroducing the stale-backup data-loss window the guard exists to prevent. Re-fetch status immediately before the import call in the `backup_url` path.
- `admin.rs:4125-4127` — on DB-update failure the error says "legacy instance restarted", which is wrong in the `backup_url` path (`restore_on_failure` deliberately does not restart a not-running instance). Make the message reflect the actual action.
- `admin.rs:4154-4256` / `4260-4358` — `parse_sse_for_backup_url` and `parse_sse_for_import_result` duplicate the stall-detection/keepalive state machine and have already drifted (only the import parser checks `event`). Extract a shared helper.
- `admin.rs:4170-4184` / `4276-4290` — the in-loop total-timeout check only fires inside the stall branch; if keepalives keep arriving, the descriptive bail (chunk count, elapsed) is unreachable and the caller gets a generic reqwest timeout instead. Total budget is still bounded by the request `.timeout(600s)`, so this is diagnostic-only — move the elapsed check to the top of each iteration.
- `admin.rs:3604-3615` — the SECRETS_MASTER_KEY preflight error tells the admin to "start all instances… before migration", contradicting the `backup_url` precondition that the instance be stopped. Branch the message.

**Checks run:** `cargo check -p api`, `cargo clippy -p api --all-targets -- -D warnings`, and `cargo fmt --check` all pass; `cargo test -p api --lib` passes (101 tests); GitHub Test Suite green on head `20e4780`. Full diff inspected plus the surrounding handler, both SSE parsers, `restore_on_failure`, and route registration. Grepped the test suite (no migrate-endpoint coverage exists) and verified no log line emits the presigned URL or customer data, per the repo's privacy rules.

Evrard-Nil (approved):

PR #307, opened 2026-06-15
fix(db): tolerate not-yet-ready members in Patroni cluster discovery

Evrard-Nil:
Mirror of **nearai/cloud-api#794** — chat-api vendors its own copy of the `database` crate, so it hit the identical failure.

## Symptom
Adding a new postgres replica logged, in chat-api **and** cloud-api:
```
ERROR database::patroni_discovery: Failed to refresh cluster state:
Failed to parse cluster info: error decoding response body
```

## Root cause (confirmed against the real struct)
Patroni lists a node as a `/cluster` member **as soon as it registers in the DCS — before it publishes its `conn_url`**, so a replica mid-creation appears **without `host`/`port`/`api_url`**. Those fields are required on `ClusterMember`, so a strict `Vec<ClusterMember>` parse fails the entire response (`missing field \`host\``) — one half-registered member poisons discovery for the whole initialization window.

## Fix
`deserialize_members` parses members individually and drops any that don't fully deserialize. `ClusterMember` stays rigid, so the rest of the discovery/selection code is untouched (surviving members all have `host`/`port`). Same philosophy as the existing `deserialize_lag` resilience.

## Tests
`9 passed; 0 failed` — adds `test_initializing_member_does_not_poison_parse` and `test_full_cluster_with_initializing_replica_parses`; fmt + clippy (`-D warnings`) clean.

**Unblocks deploying new postgres instances.**

### DIFF
--- a/crates/database/src/patroni_discovery.rs
+++ b/crates/database/src/patroni_discovery.rs
@@ -44,8 +44,43 @@ pub struct ClusterMember {
     pub timeline: Option<i64>,
 }
 
+/// Patroni lists a node as a member as soon as it registers in the DCS — which
+/// happens *before* it publishes its `conn_url`. So a replica mid-creation (or
+/// an uninitialized node) appears without `host`/`port`/`api_url`. Those fields
+/// are required on `ClusterMember`, so a strict `Vec<ClusterMember>` parse fails
+/// the ENTIRE `/cluster` response with `missing field \`host\`` over one
+/// half-registered member — freezing topology discovery for every consumer
+/// (cloud-api + chat-api) whenever a new postgres instance is being added.
+///
+/// Parse each member independently and drop any that don't fully deserialize; a
+/// member with no connection info is not a usable leader/replica target anyway,
+/// and the next refresh picks it up once it finishes registering. Same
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
The change adds `deserialize_members`, which parses each Patroni `/cluster` member independently and drops any that fail to deserialize (e.g. a replica mid-creation lacking `host`/`port`), rather than failing the whole cluster parse. This is correct and well-scoped, mirroring the existing `deserialize_lag` pattern.

- Dropping not-yet-ready members is safe: `update_cluster_state` (lines 178–192) only selects members in `running`/`streaming` state using `host`/`port`, which dropped members lack — so selection behavior is unchanged.
- The `warn!` log only emits the cluster node `name` and the serde error, exposing no customer data.
- `ClusterInfo` is deserialization-only, so the custom `#[serde(deserialize_with)]` cannot affect any serialization path. No `deny_unknown_fields`, so extra fields like `api_url` are safely ignored.

Local checks: `cargo test -p database` (9/9 passed, including the two new regression tests) and `cargo clippy -p database -- -D warnings` (clean).

think-in-universe (approved):

PR #309, opened 2026-06-15
feat: grant crabshack ownership for migrated instances

Longarithm:
## Problem

This adds **`POST /v1/admin/agents/instances/{id}/grant-owner`** (`admin_grant_instance_owner`).

After we migrate legacy instance, we should set owner on crabshack infra to enable scheduled backups.
I wouldn't do it in the process of migration, because backups can interfere with it, but instead create a separate endpoint to grant ownership, to be called once instance is fully migrated.

There is no change on user side, and the logic is straightforward, it just needs to derive user id correctly. CrabShack user id as `sha256(hex_decode(auth_secret))`, mirroring orchestrator-api's `userIdFromAuthSecret`.

## Verification

- `cargo fmt -p api` + `cargo check -p api` clean.
- `user_id` derivation cross-checked against CrabShack's `userIdFromAuthSecret` on a sample `auth_secret` — both produce the same hash, so granted owner ids match what CrabShack computes on login.
- No credentials logged (only instance id and the derived owner id, which is a hash).
- Validated live on CrabShack staging against a real owned instance: ran the endpoint's exact grant (derived owner `user_id`), after which the instance appeared as `role=owner` in the owner's own session and an owner-attributed backup completed (`ready`).

Not run: full integration suite (needs Postgres + `--features test`).

### DIFF
--- a/crates/api/src/openapi.rs
+++ b/crates/api/src/openapi.rs
@@ -102,6 +102,7 @@ use utoipa::OpenApi;
         crate::routes::admin::admin_start_instance,
         crate::routes::admin::admin_stop_instance,
         crate::routes::admin::admin_restart_instance,
+        crate::routes::admin::admin_grant_instance_owner,
         crate::routes::admin::admin_sync_agent_status,
         crate::routes::admin::bi_list_users,
         crate::routes::admin::bi_users_summary,
@@ -141,6 +142,7 @@ use utoipa::OpenApi;
         services::user::ports::AccountDeletion,
         services::user::ports::AccountDeletionStatus,
         crate::routes::admin::AdminRetryAccountDeletionResponse,
+        crate::routes::admin::GrantInstanceOwnerResponse,
         crate::models::AuthResponse,
         crate::error::ApiErrorResponse,
         // Auth request models
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -30,6 +30,7 @@ const ADMIN_CHANGE_REASON_MAX_LEN: usize = 255;
 use services::model::ports::{UpdateModelParams, UpsertModelParams};
 use services::user_usage::UsageRankBy;
 use services::UserId;
+use sha2::{Digest, Sha256};
 use std::collections::HashMap;
 use urlencoding::encode;
 use uuid::Uuid;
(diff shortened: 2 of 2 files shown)


Evrard-Nil (approved):
## Review — feat: grant CrabShack ownership for migrated instances

Reviewed the full diff (3 commits), read `crates/api/src/routes/admin.rs` and the router/middleware wiring, ran `cargo fmt --check` and `cargo check` (both clean), and checked CI (pass).

### Security

**Admin auth middleware — confirmed correct.** The route is registered inside `create_admin_router()`, which is wrapped with `admin_auth_middleware` in `routes/mod.rs:114–117`. That middleware requires both a valid session token AND that the user's email is in both `admin_domains` and `admin_emails` (or just `admin_domains` when `admin_emails` is empty — existing fallback behaviour, not introduced here). The `Extension(_user): Extension<AuthenticatedUser>` parameter is correctly declared so Axum wires the middleware-set user extension.

**`auth_secret` derivation — correct and consistent with CrabShack.** `auth_secret` is generated by `generate_random_credential(32)` = `hex::encode(32 random bytes)` = a 64-char hex string, stored encrypted in the DB. After decryption the string is still hex. `hex::decode(auth_secret.trim())` correctly re-materialises the 32 raw bytes; `hex::encode(Sha256::digest(...))` then produces the CrabShack `userIdFromAuthSecret` identity. The derivation mirrors `orchestrator-api/src/auth/user-queries.ts` as documented.

**`hex_decode` failure — handled gracefully.** Non-hex `auth_secret` maps to `ApiError::internal_server_error("Stored auth_secret is not valid hex")` — no `unwrap`/`expect`/panic. The error is logged with `instance_id` only (not the credential value).

**`auth_secret` never logged.** Confirmed: the only tracing calls in the new function log `instance_id`, `user_id`, or the derived `owner_user_id` (a SHA-256 hash = a user identity, equivalent to what's already logged at line 1970 for lifecycle actions). The `_backup_passphrase` binding on line 4187 is correctly discarded.

**CrabShack token from config, not hardcoded.** `crabshack_manager.token` comes from `AgentManager.token`, which is loaded from `AGENT_MANAGER_TOKENS` env var at startup. The `AgentManager` `Debug` impl redacts the token from logs.

**URL guard — reliable, not brittle.** `trim_end_matches('/')` normalisation is applied on both sides before comparison, matching exactly the pattern used in `admin_migrate_instance` (line 3394). `find_crabshack_manager()` looks for `url.contains("crabshack")` in the manager list — same lookup used throughout the migration flow.

**Idempotency — safe.** The PR doc says CrabShack's access grant is last-write-wins; calling the endpoint twice just re-grants the same role to the same `owner_user_id`. No unique-constraint violation, no double-grant risk. The DB read and the CrabShack POST are not atomic, but for this one-shot reconciliation use-case that's fine — a double-call produces the same end state.

**No TOCTOU concern.** There is no ownership check-then-set on chat-api's side; the only mutation is the outbound CrabShack POST, which is idempotent.

### Correctness

**UUID validation.** `Uuid::parse_str(&instance_id)` on line 4130 rejects non-UUID paths cleanly with a 400 before any DB access.

**CrabShack error surfacing.** Non-2xx responses are caught and return a 500 with the HTTP status logged. This is consistent with how `admin_migrate_instance` handles CrabShack failures — the admin caller sees a failure and the status code is in the log. The body from CrabShack is intentionally not forwarded to the caller (avoids leaking internal API details), which is appropriate for an admin-only endpoint.

**Instance name URL-encoded.** `encode(&instance.name)` (urlencoding crate) is applied in the grant URL construction — consistent with other CrabShack call sites.

**OpenAPI coverage — addressed.** The `#[utoipa::path(...)]` attribute and `openapi.rs` registration (both `admin_grant_instance_owner` path and `GrantInstanceOwnerResponse` schema) were added in the third commit (`e0509e0`), directly addressing the Gemini review comment. The `GrantInstanceOwnerResponse` correctly omits `owner_user_id` — the derived identity is not surfaced to callers, consistent with the documented privacy intent.

**CI: passing.**

### One non-blocking note

When CrabShack returns a non-2xx, the response body is consumed but not read. A 404 ("instance not found on CrabShack") and a 409 ("access already exists") both surface to the caller as the same generic 500. For a one-shot reconciliation admin tool this is acceptable — the admin can check CrabShack directly — but worth a future improvement to at least distinguish 404 from other failures.

No blocking issues. Approving.

PR #339, opened 2026-07-08
fix(webhook): reject future-dated Stripe timestamps beyond tolerance

Evrard-Nil:
Fixes #293

`StripeWebhookVerifier::verify` only rejected timestamps that were too *old*: it computed `now - header.timestamp` and rejected when that exceeded the tolerance. A future-dated (but validly signed) timestamp produced a negative age, so it slipped through, weakening replay/freshness protection.

This compares the **absolute** skew against the tolerance, so events dated too far in either direction are rejected. `checked_abs` maps the `i64::MIN` edge case to the same overflow error already emitted by `checked_sub`, so the existing overflow test still holds. The test that previously asserted a future timestamp was accepted now asserts rejection, plus a new test confirms a future timestamp *within* tolerance is still accepted.

Scope note: this addresses finding #1 (timestamp validation) of the issue. Finding #2 (narrowing the DB transaction scope around outbound Stripe API calls in `SubscriptionServiceImpl::handle_stripe_webhook`) is a larger, separate refactor and is intentionally left as a follow-up.

### DIFF
--- a/crates/stripe_client/src/webhook.rs
+++ b/crates/stripe_client/src/webhook.rs
@@ -48,11 +48,17 @@ impl StripeWebhookVerifier {
             return Err(StripeWebhookError::SignatureMismatch);
         }
 
-        let timestamp_age = now
+        // Reject timestamps outside the tolerance in *both* directions. Comparing the
+        // absolute skew ensures a validly signed but future-dated timestamp (excessive
+        // clock skew or a replayed event) beyond tolerance is rejected, not just stale
+        // ones. `checked_abs` maps the `i64::MIN` edge case to the same overflow error
+        // already produced by `checked_sub`.
+        let timestamp_skew = now
             .timestamp()
             .checked_sub(header.timestamp)
+            .and_then(i64::checked_abs)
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
Small, correct, well-contained fix. `webhook.rs:56-63` now computes the signed skew via `checked_sub` then `checked_abs`, so a validly-signed but future-dated timestamp beyond tolerance is rejected symmetrically — closing the freshness/replay gap from #293. Overflow safety is preserved (`checked_sub(i64::MIN)` short-circuits to `TimestampOutsideTolerance`, and the existing `rejects_timestamp_age_overflow` test still passes), and the boundary is handled correctly (`>` not `>=`: skew 301 rejected, 300 accepted). The call site in `stripe_client_adapter.rs` only maps success/failure, so there's no downstream impact.

No blocking issues found.

Checks: `cargo test -p stripe_client` passed (14/14, including the new future-timestamp tests and the preserved overflow test); `cargo fmt --check` and `cargo clippy -- -D warnings` clean. CI `cargo deny`, `CodeQL (actions)`, and `claude-review` passed; `Test Suite` and `CodeQL (rust)` were still running at review time.

PR #341, opened 2026-07-08
feat: Lukka account risk checks for HoS NEAR wallet flows

think-in-universe:
## Summary
- add configurable Lukka KYT client for NEAR AML score checks with caching, timeout/retry handling, redacted config debug output, and normalized `KytCheckResult` responses
- run KYT checks before House-of-Stake NEAR wallet subscription, cancel, resume, plan change, and credit purchase intents
- surface KYT results in backend HoS responses while keeping provider failures as `UNKNOWN` with a reason
- document Lukka env vars in `env.example` and Docker Compose pass-through

Closes nearai/house-of-stake-contracts#55

## Validation
- `cargo fmt`
- `cargo check -p services`
- `cargo check -p api`
- `cargo test -p services kyt`
- `cargo test -p api --test subscriptions_tests test_change_plan_outcome_serde_uses_kind_discriminant`


think-in-universe (commented):
Some comments: 

1. I think we probably need to trace the high risk accounts in one database table with all the details. so we can audit the accounts and transactions in the future.
2. we should also send Slack notification via webhook for any high risk account detection.

crates/api/src/routes/users.rs
```diff
@@ -6,9 +6,55 @@ use axum::{
...
+pub async fn get_user_status(
+    State(app_state): State<AppState>,
+    Extension(user): Extension<AuthenticatedUser>,
+) -> Result<Json<UserStatusResponse>, ApiError> {
+    app_state
+        .user_service
+        .check_user_status(user.user_id)
+        .await
+        .map_err(|e| match e {
+            UserStatusError::AmlHighRiskBlocked { .. } => ApiError::forbidden("Account blocked"),
```
think-in-universe: Maybe return a more neutral message like `Account error`

crates/services/src/subscription/service.rs
```diff
@@ -79,6 +83,7 @@ const TTL_CACHE_SECS: u64 = 600; // 10 minutes
 const SYSTEM_CONFIGS_TTL_CACHE_SECS: u64 = 60; // 1 minute
 /// Upper bound for cached HoS stake sources; invalidate_credit_limit_cache also clears this on subscription, credit, and sync changes.
 const HOUSE_OF_STAKE_STAKE_SOURCE_TTL_SECS: u64 = 60;
+const AML_REPORT_REFRESH_DAYS: i64 = 30;
```
think-in-universe: Let's make the refresh days configurable via env vars.

env.example
```diff
@@ -121,6 +121,19 @@ POSTHOG_HOST=https://us.i.posthog.com
 STRIPE_SECRET_KEY=
 STRIPE_WEBHOOK_SECRET=
 
+# Lukka AML checks for NEAR wallet billing flows
+LUKKA_AML_ENABLED=false
+LUKKA_BASE_URL=https://api.blockchain-analytics.lukka.tech
+LUKKA_BEARER_TOKEN=
+LUKKA_API_KEY=
```
think-in-universe: Do we really need `LUKKA_API_KEY` and `LUKKA_API_SECRET`? I think `LUKKA_BEARER_TOKEN` is enough.

crates/api/src/main.rs
```diff
@@ -142,7 +146,15 @@ async fn main() -> anyhow::Result<()> {
...
 
-    let user_service = Arc::new(UserServiceImpl::new(user_repo.clone()));
+    let user_service = Arc::new(UserServiceImpl::new_with_aml_alerts(
+        user_repo.clone(),
+        aml_service.clone(),
+        aml_report_repo.clone(),
+        config.lukka_aml.high_risk_slack_webhook_url.clone(),
+        config.lukka_aml.high_risk_slack_timeout_ms,
+        config.lukka_aml.report_refresh_days,
+        config.lukka_aml.high_risk_slack_alert_on_cached_reports,
```
think-in-universe: For LINE 153-157, can we make `send_high_risk_aml_slack_alert` a function of `aml_service`, so we don't need to pass all the configs separately?

crates/database/src/migrations/sql/V36__index_aml_reports_created_at.sql
```diff
@@ -0,0 +1,4 @@
+CREATE INDEX idx_aml_risk_reports_created_at ON aml_risk_reports(created_at DESC);
```
think-in-universe: Can we merge V36 into V35 if possible?

hanakannzashi (approved):

PR #344, opened 2026-07-14
fix: preserve canceled HoS subscriptions

think-in-universe:
## Summary

Preserve House of Stake subscription history when the on-chain subscription is no longer active.

Fixes nearai/house-of-stake-contracts#44.

## Root Cause

`GET /v1/subscriptions?include_inactive=true` is DB-only, but HoS sync deleted local `house-of-stake` rows when the contract probe returned `null`. After an effective cancellation, there was no local row left for `include_inactive=true` to return, unlike Stripe canceled subscriptions.

## Changes

- Mark expired HoS rows as `canceled` during entitlement checks instead of deleting them.
- Mark existing local HoS rows as `canceled` when chain sync returns no subscription instead of deleting them.
- Clear pending downgrade fields when preserving the canceled HoS row.
- Update HoS subscription tests to assert the preserved inactive row is returned by `include_inactive=true`.

## Validation

- `cargo test -p api --features test --test subscriptions_tests test_near_staking_sync_marks_local_hos_canceled_when_chain_returns_null`
- `cargo test -p api --features test --test subscriptions_tests test_near_staking_sync_falls_back_to_configured_prices_before_canceling_local_row`
- `cargo test -p api --features test --test subscriptions_tests test_near_staking_sync_marks_expired_cancel_at_period_end_row_canceled`
- `cargo test -p api --features test --test subscriptions_tests test_cancel_subscription_house_of_stake_returns_wallet_intent_message`

### DIFF
--- a/crates/api/src/routes/subscriptions.rs
+++ b/crates/api/src/routes/subscriptions.rs
@@ -677,7 +677,7 @@ pub async fn create_portal_session(
     path = "/v1/subscriptions/near/sync",
     tag = "Subscriptions",
     responses(
-        (status = 200, description = "Reconcile finished; see `skipped`, `deleted_house_of_stake_rows`, `upserted_house_of_stake_row`, and optional `skipped_reason` in the body.", body = NearStakingSyncSummary),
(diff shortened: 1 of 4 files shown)


hanakannzashi (approved):
Reviewed the latest head (aaa2d2e). The HoS cancellation paths now preserve history consistently, only transition active/trialing rows, clear pending downgrade state, and avoid re-touching canceled history. The sync summary/API updates are coherent, regression coverage is solid, and all checks pass. LGTM.

PR #347, opened 2026-07-16
fix(database): follow the Patroni leader across failovers

PierreLeGuen:
Mirror of nearai/cloud-api#887.

## Problem

After the 2026-07-12 Patroni leader failover, chat-api stayed connected to the failed leader for ~30h (IronClaw auth broken, `Timeout occurred while creating a new object` pool errors from chat-api and chat-api-worker) until a manual restart. The failover machinery exists but is dead wiring: repositories, the task worker, and the subscription service all hold clones of the **startup** deadpool `Pool` (target host fixed at creation), and the ClusterManager's leader-change handling swaps a pool object nothing re-reads. A failed swap was also never retried, and an in-recovery leader was installed with only a warning.

## Fix

- **`DbPool` is now a shared, swappable handle**: the ClusterManager installs the new leader's pool into it on failover and every clone — repositories, task worker, subscription service — immediately routes new acquisitions to the new leader. Call sites are unchanged (`.get().await` as before).
- The handle type lives in the **services** crate (`services::db_pool::DbPool`, re-exported as `database::pool::DbPool`) because `SubscriptionServiceImpl` holds it directly and `database` depends on `services`; its `db_pool` field previously held a raw `deadpool_postgres::Pool`, which would have stayed pinned to the old leader even with the swap in place.
- **Reconcile loop** replaces the one-shot leader tracking: the manager records which `host:port` the write pool actually targets (only on a successful, verified build) and converges on the discovered leader every 30s tick, retrying failed rebuilds with the error cause logged.
- **Leader verification is bounded** (15s timeout) and an in-recovery leader is rejected (retried next tick) instead of installed as a read-only write pool.
- Stale discovery state logs a warning; read paths no longer hold a lock across pool acquisition or divide by zero on an empty replica list.

## Tests

- `failed_write_pool_rebuild_is_retried_on_next_reconcile`: a failed rebuild must not record the leader as installed and must be retried.
- `reconcile_repoints_startup_pool_clones_to_new_leader` (runs against the CI Postgres): a pool handle cloned at startup must serve connections to the new leader after reconcile.
- `replace_propagates_to_existing_clones` / `get_on_uninitialized_handle_fails_closed`.
- Full suite green locally except two pre-existing `agent_tests` failures (500 vs 402) that fail identically on unmodified origin/main with the same local env.

### DIFF
--- a/crates/database/src/cluster_manager.rs
+++ b/crates/database/src/cluster_manager.rs
@@ -137,11 +137,8 @@ impl ClusterManager {
             // as leader while Postgres is still completing promotion;
             // installing it would pin every write to a read-only node with no
             // retry, so fail and let the next reconcile tick try again.
-            let rows = conn.query("SELECT pg_is_in_recovery()", &[]).await?;
-            let is_replica: bool = rows
-                .first()
-                .ok_or_else(|| anyhow!("Empty response to pg_is_in_recovery()"))?
-                .try_get(0)?;
+            let row = conn.query_one("SELECT pg_is_in_recovery()", &[]).await?;
+            let is_replica: bool = row.try_get(0)?;
             if is_replica {
                 return Err(anyhow!(
                     "Node {} claims to be leader but is still in recovery",
@@ -318,10 +315,13 @@ impl ClusterManager {
     /// dropped.
     pub async fn reconcile(&self) {
         if self.discovery.is_state_stale().await {
-            warn!(
-                "Patroni cluster state is stale (age: {:?}s); leader changes may go undetected",
-                self.discovery.get_state_age_secs().await
-            );
+            let age = self
+                .discovery
+                .get_state_age_secs()
+                .await
+                .map(|secs| format!("{secs}s"))
+                .unwrap_or_else(|| "never loaded".to_string());
+            warn!("Patroni cluster state is stale (age: {age}); leader changes may go undetected");
         }
 
         match self.discovery.get_leader().await {
--- a/crates/database/src/cluster_manager.rs
+++ b/crates/database/src/cluster_manager.rs
@@ -1,5 +1,5 @@
 use crate::patroni_discovery::{ClusterMember, PatroniDiscovery};
-use crate::pool::create_pool_with_native_tls;
+use crate::pool::{create_pool_with_native_tls, DbPool};
 use anyhow::{anyhow, Result};
 use deadpool::managed::QueueMode;
 use deadpool_postgres::{Config, Object as PooledConnection, Pool, Runtime};
@@ -10,6 +10,12 @@ use std::time::Duration;
 use tokio::sync::RwLock;
 use tokio::time;
 use tracing::{debug, error, info, warn};
+
+/// Upper bound on verifying a candidate leader before installing its pool.
+/// Generous next to the 5s pool create/wait timeouts it wraps, but bounded so
+/// a member that accepts connections and then hangs cannot stall the
+/// reconcile loop.
+const WRITE_POOL_VERIFY_TIMEOUT: Duration = Duration::from_secs(15);
 
 /// Read preference strategy
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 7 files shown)


crates/database/src/cluster_manager.rs
```diff
@@ -106,24 +124,44 @@ impl ClusterManager {
...
+            .await
+            .map_err(|_| {
+                anyhow!(
+                    "Timed out verifying leader {} after {:?}",
+                    leader.name,
+                    WRITE_POOL_VERIFY_TIMEOUT
+                )
+            })??;
+
+        self.write_pool.replace(pool);
```
lloydmak99: **Medium — stale candidate can be installed after discovery advances.** `reconcile()` snapshots a leader, but verification can take up to 15 seconds. Patroni discovery refreshes concurrently, so a newer leader may be known by the time this unconditional `replace()` runs. The stale candidate can then repoint every repository until the next 30-second reconcile. Re-read discovery immediately before installation and require its current `host:port` to still match the verified candidate; also serialize reconciliation so concurrent callers cannot produce last-writer-wins rollback.

crates/database/src/cluster_manager.rs
```diff
@@ -369,6 +398,157 @@ pub struct ClusterStats {
...
+            test_db_config(),
+            ReadPreference::LeaderOnly,
+            None,
+        );
+
+        // What Database::new does at startup: take a handle and clone it into
+        // the repositories.
+        let repository_handle = manager.write_pool();
+
+        manager.reconcile().await;
```
lloydmak99: **Medium — this does not reproduce replacement of an active old-leader pool.** The handle starts uninitialized, the dead-leader reconcile installs nothing, and this assertion confirms that state. The later reconcile therefore performs the first usable installation rather than replacing a pool that repository clones were already using. Seed a usable old pool, clone the handle and acquire through it, reconcile to a distinguishable new pool, then verify that the same clone acquires from the new target.

lloydmak99 (changes requested):
Two mirrored issues need addressing before merge:

1. The candidate leader can become stale during the bounded verification window and still be installed afterward.
2. The regression test covers first installation into an uninitialized handle, not replacement of an already-serving old-leader pool.

All current chat-api checks pass.

crates/database/src/cluster_manager.rs
```diff
@@ -106,24 +129,56 @@ impl ClusterManager {
...
+        if current.as_ref().map(Self::member_target).as_deref() != Some(target.as_str()) {
+            return Err(anyhow!(
+                "Cluster leader changed to {:?} while verifying {target}; discarding this pool",
+                current.map(|m| Self::member_target(&m))
+            ));
         }
 
-        let mut write_pool = self.write_pool.write().await;
-        *write_pool = Some(pool);
+        self.write_pool.replace(pool);
```
lloydmak99: **Medium — the stale-leader race is narrowed but not eliminated.** `get_leader().await` releases the discovery-state read lock before this `replace()`. The independent discovery refresh can publish leader B after the equality check but before stale candidate A is installed; `reconcile_lock` only serializes reconcile callers, not discovery updates. Make comparison and installation atomic relative to topology publication, for example by retaining a discovery read guard through synchronous installation or using a topology generation with coordinated compare-and-install.

crates/database/src/cluster_manager.rs
```diff
@@ -369,6 +419,280 @@ pub struct ClusterStats {
...
+        ));
+        let repository_handle = manager.write_pool();
+
+        let reconcile_task = tokio::spawn({
+            let manager = manager.clone();
+            async move { manager.reconcile().await }
+        });
+        // Wait until reconcile is inside A's delayed verification, then fail
+        // over.
+        tokio::time::sleep(Duration::from_millis(250)).await;
```
lloydmak99: **Medium — this fixed sleep makes the regression test scheduler-racy.** It does not prove the spawned reconcile task has snapshotted or connected to leader A. Under CI contention, reconcile may start only after discovery changes to B, legitimately install B, and fail the later assertion that no pool was installed. Synchronize on leader A accepting its first connection (or use a oneshot/barrier from the proxy) before changing discovery.

lloydmak99 (changes requested):
Re-review of the current head:

- The previous active-pool replacement test gap is resolved: the new test proves an existing repository handle moves from leader A to B and keeps working after A is stopped.
- Two mirrored issues remain: the final revalidation/install sequence is still TOCTOU, and the new stale-candidate regression test relies on a scheduler-sensitive fixed sleep.

All current chat-api checks pass.

lloydmak99 (approved):
## Code review

Approved at `fc704a5d27e829f64d89fbf8d62ac2f4a3417b99`.

I reviewed the mirrored failover implementation at the current head, including the services-owned shared handle required by the existing crate dependency direction, subscription-service usage, repository call sites, discovery locking, and retry paths. The previous compare/install TOCTOU and scheduler-dependent test findings are resolved by the same atomic discovery-guarded install and accepted-connection synchronization as cloud-api.

Validation:
- Local shared-handle replacement tests: 2 passed.
- Local failed-rebuild retry regression passed.
- The two PostgreSQL-backed failover tests could not run locally because this host has no PostgreSQL listener or Docker, but both exact tests passed in the current CI Test Suite.
- Test Suite, CodeQL, and cargo-deny are green.

No blocking findings. Non-blocking future coverage could directly exercise a candidate returning `pg_is_in_recovery() = true` and two simultaneous `reconcile()` callers.

think-in-universe (approved):

PR #358, opened 2026-07-28
fix: use effective HoS locks for credits

think-in-universe:
## Summary

Updates House of Stake credit and plan-change reads to request the contract's effective lock view.

This lets chat-api calculate stake-based credits from the projected post-downgrade stake once a scheduled downgrade is ready, instead of using stale stored `get_lock().amount_near`.

## Root Cause

`/v1/subscriptions/near/sync` reads projected subscription views, so the local subscription row can show the downgraded plan. Stake-based credits then separately call `get_lock(last_lock_id)` and use raw `amount_near`, which remains at the old stake until a state-changing contract mutation applies the pending update.

## Changes

- Added `view_get_effective_lock`, sending `effective: true` to HoS `get_lock`.
- Switched HoS stake-based credits to use effective lock amount.
- Switched HoS change-plan current stake reads to use effective lock amount.
- Added test coverage that verifies `effective: true` is sent and credits use the effective 10 NEAR value.

## Validation

- `cargo fmt -p services -p api`
- `cargo test -p services subscription::near_staking::tests -- --nocapture`
- `cargo check -p services -p api`

Attempted DB-backed API regression test:

- `cargo test -p api --test subscriptions_tests test_house_of_stake_credits_use_effective_lock_amount_after_projected_downgrade -- --nocapture`
- Blocked locally by Postgres auth: `FATAL: password authentication failed for user "postgres"`.
- The test target compiled before failing during database migration setup.

Related to https://github.com/nearai/house-of-stake-contracts/issues/143

### DIFF
--- a/crates/services/src/subscription/service.rs
+++ b/crates/services/src/subscription/service.rs
@@ -1,7 +1,8 @@
 use super::near_staking::{
-    lock_amount_yocto, subscription_row_from_chain, view_get_effective_lock, view_get_price,
-    view_get_purchase, view_get_subscription_for_price, view_storage_balance_bounds,
-    view_storage_balance_of, NearStakingStorageBalance, NearStakingStorageBalanceBounds,
+    lock_amount_yocto, subscription_row_from_chain, view_get_effective_lock, view_get_lock,
+    view_get_price, view_get_purchase, view_get_subscription_for_price,
+    view_storage_balance_bounds, view_storage_balance_of, NearStakingStorageBalance,
+    NearStakingStorageBalanceBounds,
 };
 use super::ports::{
     BillingCycleAnchor, BillingPeriod, CancelSubscriptionOutcome, ChangePlanOutcome,
@@ -2515,7 +2516,7 @@ impl SubscriptionService for SubscriptionServiceImpl {
                 .ok_or_else(|| {
                     SubscriptionError::InternalError("HoS subscription missing last_lock_id".into())
                 })?;
-            let lock_j = view_get_effective_lock(&self.near_rpc_url, contract_id, last_lock_id)
+            let lock_j = view_get_lock(&self.near_rpc_url, contract_id, last_lock_id)
                 .await
                 .map_err(Self::near_rpc_err)?
                 .ok_or_else(|| {
--- a/crates/services/src/subscription/service.rs
+++ b/crates/services/src/subscription/service.rs
@@ -1,8 +1,8 @@
 use super::near_staking::{
-    lock_amount_yocto, subscription_row_from_chain, view_get_effective_lock, view_get_lock,
-    view_get_price, view_get_purchase, view_get_subscription_for_price,
-    view_storage_balance_bounds, view_storage_balance_of, NearStakingStorageBalance,
-    NearStakingStorageBalanceBounds,
+    lock_amount_yocto, lock_has_active_shares, subscription_row_from_chain,
+    view_get_effective_lock, view_get_lock, view_get_price, view_get_purchase,
+    view_get_subscription_for_price, view_storage_balance_bounds, view_storage_balance_of,
+    NearStakingStorageBalance, NearStakingStorageBalanceBounds,
 };
 use super::ports::{
     BillingCycleAnchor, BillingPeriod, CancelSubscriptionOutcome, ChangePlanOutcome,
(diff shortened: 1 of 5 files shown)


crates/services/src/subscription/service.rs
```diff
@@ -2514,7 +2515,7 @@ impl SubscriptionService for SubscriptionServiceImpl {
                 .ok_or_else(|| {
                     SubscriptionError::InternalError("HoS subscription missing last_lock_id".into())
                 })?;
-            let lock_j = view_get_lock(&self.near_rpc_url, contract_id, last_lock_id)
+            let lock_j = view_get_effective_lock(&self.near_rpc_url, contract_id, last_lock_id)
```
hanakannzashi: [P2] Do not turn an unapplied due decrease into a no-op

After a scheduled 50→25 NEAR decrease is due, this returns the effective 25 while the stored lock is still 50 and pending unstake is still zero. If the user submits the same plan with `target_amount = 25`, the check below sees `target_amount == current_lock_amount` and returns `NoOp`, so no wallet `update_subscription` call is emitted to apply the due update. Before this change the raw 50 differed from the target and produced the required 1-yocto intent; the contract lifecycle test also commits a due decrease by calling `update_subscription` again with that same target. Please retain enough raw/effective state to skip the NoOp path whenever they differ.

hanakannzashi (commented):
[P2] Do not use the view-time projection as an exact payable quote

The contract ordering does apply the due update before computing the new stake delta, so using the lower effective amount is directionally correct. However, `get_lock(effective=true)` projects from the validator balance cached at view time, while `update_subscription` first refreshes the pool balance/settles and then requires `deposit == expected_delta` exactly. A reward refresh can change the rounded projection by only 1 yocto: a view can quote effective 10 NEAR and a 40 NEAR deposit for target 50, while settlement leaves `10 NEAR + 1 yocto` and the callback accepts only `40 NEAR - 1 yocto`; the wallet transaction then reverts. Effective view is appropriate for credits, but change-plan needs an atomic quote/apply path or contract-side bounded overpayment/refund before this can safely drive an exact payable amount.

crates/services/src/subscription/service.rs
think-in-universe: Fixed in bfa0606 by moving HoS `change_plan` back to the stored lock read (`view_get_lock`). `effective=true` is now only used for stake-based credits, so due raw/effective differences no longer collapse this path into `NoOp`, and the wallet intent/deposit quote stays based on the stored lock amount.

I also tightened the change-plan mock to fail if `get_lock` is called with `effective`, so the cancel-pending-downgrade path covers this regression.

crates/services/src/subscription/service.rs
```diff
@@ -1103,28 +1104,29 @@ impl SubscriptionServiceImpl {
...
-                    subscription_id = %subscription.subscription_id,
-                    last_lock_id,
-                    error = %err,
-                    "cannot resolve HoS stake-based credits: lock RPC failed"
-                );
-                return None;
-            }
-        };
+        let lock =
+            match view_get_effective_lock(&self.near_rpc_url, contract_id, last_lock_id).await {
```
hanakannzashi, line 1108: [P2] Reject effective locks that have exited all shares

The effective-lock response cannot be treated as an amount-only entitlement source. House-of-Stake #156 explicitly supports a full projected share exit: for example, a nominal 50-NEAR lock whose shares are worth 20 NEAR and a due target of 25 NEAR returns amount_near=30 NEAR, shares=0, and status=UnlockRequested. NearStakingLock currently deserializes only amount_near, so this path grants credits for 30 NEAR even though the lock has no remaining stake; the subscription itself can still be Active, so the earlier subscription check does not reject it. Please deserialize status and shares and grant stake-based credits only for an Active lock with nonzero shares. The inactive/full-exit case must return an explicit zero-credit result rather than None, because None currently falls back to the configured static plan allowance and remains fail-open. Add a regression response with nonzero amount_near plus shares=0/status=UnlockRequested.

think-in-universe: Fixed in ad13384.

Changes:
- `NearStakingLock` now deserializes `status` and `shares` in addition to `amount_near`.
- HoS stake-based credits now grant from an effective lock only when it is explicitly `Active` with `shares > 0`.
- A non-active / zero-share effective lock returns explicit zero stake-based credits and caches that zero source, so it does not fall through to the static plan allowance.
- Added unit coverage for the active-share gate and an API regression response with `amount_near = 30 NEAR`, `shares = 0`, `status = UnlockRequested`, asserting `plan_credits == 0`.

Validation:
- `cargo fmt -p services -p api`
- `cargo test -p services subscription::near_staking::tests -- --nocapture`
- `cargo check -p services -p api`

The two focused API regression filters compile but local execution still stops at DB migration setup because Postgres auth fails for user `postgres`.

hanakannzashi (approved):

PR #363, opened 2026-07-29
feat: prorate HoS stake credit entitlement

### LINKED ISSUE #132: Prorate stake-based monthly credits for mid-period stake increases
## Problem

House-of-Stake-backed monthly credits should track the actual staking reward window, not just the latest requested stake amount.

For chat-api/private-assistant, the risky case is a user increasing staked NEAR near the end of a subscription period and then canceling. Cancellation does not shorten the subscription period, but granting the full higher monthly credit limit immediately would let the user consume more credits than the staking rewards attributable to the increased stake for that period.

Decreases have the opposite property: reducing the requested stake does not trigger unstaking until the subscription period ends, so the current period's actual staking rewards do not decrease. The current period credit limit should therefore remain unchanged for stake decreases and only drop in the next period when the lower stake becomes effective.

## Proposed policy

For stake-based credit plans:

1. Cancellation does not alter the current subscription period.
2. Stake decreases do not reduce current-period monthly credits. They affect the next period after the lower stake is effective.
3. Stake increases grant only prorated additional credits for the remaining time in the existing subscription period.
4. The next period starts fresh from the stake amount effective at that new period start.

Prorated increase formula:

```text
base_monthly_credits = stake_at_period_start * credits_per_staked_near

if effective_stake increases during the period:
  added_credits =
    (new_effective_stake - previously_credited_stake)
    * credits_per_staked_near
    * (current_period_end - increase_effective_at)
    / (current_period_end - current_period_start)

  current_period_credit_limit += added_credits
  previously_credited_stake = new_effective_stake

if effective_stake decreases or a decrease is pending:
  current_period_credit_limit remains unchanged
```

Example:

```text
Period: 30 days
Rate: $0.01 credits per staked NEAR per month
Stake at period start: 500 NEAR => $5 current-period credits

Day 20: user increases to 800 NEAR
Delta stake: 300 NEAR
Full-month delta: $3
Remaining period: 10 / 30

Prorated current-period increase: $3 * 10 / 30 = $1
Current-period credit limit: $6
Next period, if 800 NEAR remains effective: $8
```

## Implementation notes

Both chat-api and private-assistant should avoid deriving the current-period entitlement only from the latest lock amount. They need a per-period entitlement snapshot, for example:

```text
user_id
subscription_id
period_start
period_end
credited_stake_yocto
credit_limit_nano_usd
last_reconciled_at
```

On reconcile / credit-limit calculation:

1. If no snapshot exists for the active period, create one using the currently effective stake.
2. If the effective on-chain stake is greater than `credited_stake_yocto`, add prorated credits for only the delta stake and update `credited_stake_yocto`.
3. If the effective on-chain stake is less than or equal to `credited_stake_yocto`, keep `credit_limit_nano_usd` unchanged for the current period.
4. When the next period starts, create a new snapshot from the then-effective stake.

## Contract/API consideration

The consumers need enough House-of-Stake contract data to distinguish:

- current effective stake / lock amount,
- pending stake-only updates,
- update apply time,
- subscription period start/end or enough data to derive the current period,
- whether cancellation changes only renewal state and not the active period.

If any of that data is missing or ambiguous from current view methods, please add/adjust view output so chat-api and private-assistant can implement this deterministically.

## Acceptance criteria

- chat-api and private-assistant apply stake decreases only from the next effective period.
- chat-api and private-assistant prorate stake increases by remaining time in the current period.
- Repeated stake increases within one period only prorate the incremental delta each time.
- Canceling after a late-period stake increase does not grant a full-month credit increase.
- Tests cover late-period increase + cancel, repeated increases, pending decrease, and next-period reset.

think-in-universe:
## Summary

Implements per-period House-of-Stake stake-credit entitlement snapshots for variable stake-backed plans.

- Adds a `house_of_stake_credit_entitlement_snapshots` table keyed by subscription and billing period.
- Seeds each period from the effective stake, prorates only incremental stake increases by remaining period time, and keeps current-period credits unchanged for decreases.
- Uses HoS `start_ns`/`end_ns` when available for the billing period and bypasses cached HoS plan limits so repeated stake changes reconcile immediately.
- Adds regression coverage for late-period increase plus cancel, repeated increase, current-period decrease behavior, and next-period reset.

Fixes https://github.com/nearai/house-of-stake-contracts/issues/132

## Validation

- `cargo fmt`
- `cargo check -p services`
- `cargo check -p database`
- `cargo check -p api`
- `cargo test -p api --test subscriptions_tests test_house_of_stake_credit_snapshot_prorates_increases_and_resets_next_period --no-run`

Note: Running the DB-backed integration test failed locally because the configured Postgres credentials are rejected (`password authentication failed for user "postgres"`). The test target compiles successfully.


hanakannzashi (dismissed):

crates/services/src/subscription/service.rs
```diff
@@ -2007,6 +2487,70 @@ fn generate_credit_checkout_idempotency_key(
...
+                let snapshot_limit = snapshot_limit.min(i64::MAX as u64) as i64;
+                if snapshot_limit > plan_credits {
+                    tracing::warn!(
+                        user_id = %user_id.0,
+                        subscription_id = %subscription.subscription_id,
+                        plan_credits,
+                        snapshot_limit,
+                        "using current HoS snapshot as purchased-credit reconciliation floor after zero entitlement"
+                    );
+                    return Ok(Some((snapshot_limit, snapshot_start, snapshot_end)));
```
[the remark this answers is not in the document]
think-in-universe: Addressed in 7b80dfb. The resolver now carries HoS entitlement source metadata. Confirmed-zero chain results bypass the snapshot floor and purchased-credit reconciliation runs against a zero plan; snapshot fallback remains limited to transient or unknown failures. The test now asserts confirmed chain absence charges purchased credits against zero.

crates/services/src/subscription/service.rs
```diff
@@ -977,36 +1132,272 @@ impl SubscriptionServiceImpl {
...
+                    ON CONFLICT (subscription_id, period_start, period_end) DO NOTHING
+                    RETURNING credited_stake_yocto, credit_limit_nano_usd
+                    "#,
+                    &[
+                        &user_id,
+                        &subscription_id,
+                        &period_start,
+                        &period_end,
+                        &effective_stake_yocto.to_string(),
+                        &(full_period_limit as i64),
```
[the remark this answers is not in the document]
think-in-universe: Addressed in 7b80dfb. Missing HoS snapshot rows now seed at zero and only add the currently observed stake as a prorated remaining-period delta, so a first API call after a late stake increase cannot receive the full-period allowance. I also updated the regression coverage around first-observed stake proration.

crates/services/src/subscription/service.rs
```diff
@@ -1087,9 +1520,34 @@ impl SubscriptionServiceImpl {
...
+                "subscription_not_active_on_chain",
+            ));
         }
 
+        let (period_start, period_end) = match chain_sub
+            .start_ns
+            .and_then(Self::ns_to_datetime)
+            .zip(Self::ns_to_datetime(chain_sub.end_ns))
+        {
+            Some((start, end)) if end > start => (start, end),
```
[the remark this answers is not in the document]
think-in-universe: Addressed in 7b80dfb. Successful HoS resolution now uses the persisted local billing period key rather than switching to chain start_ns, so RPC fallback and recovery use the same period identity. The tests align seeded snapshots with the local period key and keep current snapshot fallback coverage.

crates/services/src/subscription/service.rs
```diff
@@ -79,10 +77,32 @@ struct CachedSystemConfigs {
...
+    period_start: chrono::DateTime<Utc>,
+    period_end: chrono::DateTime<Utc>,
+    source: HouseOfStakeEntitlementSource,
+}
+
 const TTL_CACHE_SECS: u64 = 600; // 10 minutes
+const HOUSE_OF_STAKE_CREDIT_LIMIT_CACHE_SECS: u64 = 60; // 1 minute
 const SYSTEM_CONFIGS_TTL_CACHE_SECS: u64 = 60; // 1 minute
-/// Upper bound for cached HoS stake sources; invalidate_credit_limit_cache also clears this on subscription, credit, and sync changes.
-const HOUSE_OF_STAKE_STAKE_SOURCE_TTL_SECS: u64 = 60;
```
think-in-universe, line 85: Why do we remove this source TTL for Hos?

crates/services/src/subscription/service.rs
```diff
@@ -116,8 +136,8 @@ pub struct SubscriptionServiceImpl {
     near_rpc_url: String,
     near_network_id: String,
     credit_limit_cache: Arc<RwLock<HashMap<UserId, CachedCreditLimit>>>,
-    house_of_stake_stake_source_cache:
-        Arc<RwLock<HashMap<(UserId, String), CachedHouseOfStakeStakeSource>>>,
+    house_of_stake_refresh_failure_cache:
```
think-in-universe, line 140: Why do we need this refresh failure cache, but remove `CachedHouseOfStakeStakeSource`?

crates/services/src/subscription/service.rs
```diff
@@ -3341,28 +3963,23 @@ impl SubscriptionService for SubscriptionServiceImpl {
             .get_active_subscription_for_entitlement(user_id)
             .await?;
 
-        // 1. Get plan credits (monthly_credits) from the config. Cache 10 mins.
+        // 1. Get plan credits (monthly_credits) from the config. Cache briefly for HoS because
```
think-in-universe, line 4177: 1 minute cache is too short for credit limit. 

Can we query NEAR RPC for `resolving stake-based credits` only when someone updates the subscriptions?

crates/services/src/subscription/service.rs
```diff
@@ -79,10 +77,32 @@ struct CachedSystemConfigs {
     cached_at: Instant,
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum HouseOfStakeEntitlementSource {
```
think-in-universe, line 81: Why do we need these HoS sources?

crates/services/src/subscription/service.rs
think-in-universe: They protect different paths. `CachedHouseOfStakeStakeSource` cached successful stake-source resolution, which is now folded into `credit_limit_cache` as the resolved HoS entitlement. `house_of_stake_refresh_failure_cache` is a negative/backoff cache for a different problem: when the local HoS row is expired and NEAR RPC is down, `get_active_subscription_for_entitlement` runs before the credit-limit cache can help. Without this backoff, every request retries the expired-row sync and can block on NEAR RPC. So we removed the duplicate success cache but kept a small failure cache for the pre-entitlement refresh path.

crates/services/src/subscription/service.rs
think-in-universe: We removed the separate HoS stake-source TTL because it was duplicating the credit-limit cache but did not carry the full data the hot paths need now. After proration/snapshot changes, callers need the resolved plan limit, billing period, and entitlement source together. Keeping only `credit_limit_cache` avoids two caches getting out of sync and lets both proxy admission and usage reconciliation share the same short-lived resolved entitlement.

crates/services/src/subscription/service.rs
think-in-universe: Yes, that would be the cleaner long-term direction: make stake-based entitlement update only on subscription sync / chain mutation / renewal jobs, and have request paths read the persisted snapshot only. I kept the 60s cache here because this PR preserves the current lazy reconciliation behavior while preventing per-request RPC and row locks. Moving entirely to sync-only would need one more design change: a reliable way to seed/refresh snapshots at every HoS renewal and stake change, plus a background or webhook-driven sync path so credit limits do not stay stale when users update stake on-chain without calling our sync endpoint. I can follow up with that refactor, but for this PR the short TTL is the guardrail around the remaining lazy path.

crates/services/src/subscription/service.rs
think-in-universe: The source is needed because the same numeric plan value can mean different things for purchased-credit reconciliation. A `ConfirmedZero` means the chain definitively says the subscription/lock is gone or inactive, so purchased credits must reconcile against zero. A `TransientFallback` means RPC/config/account lookup was inconclusive, so we may use a current snapshot or skip reconciliation rather than irreversibly charging purchased credits during an outage. `Chain` is the normal successful entitlement. Carrying this source through the cache prevents later code from treating confirmed-zero and transient fallback as the same `0`.

think-in-universe:
@ironloopai review

think-in-universe:
@claude review

think-in-universe:
@claude review

think-in-universe:
@ironloopai review

hanakannzashi (dismissed):

think-in-universe:
@ironloopai review

think-in-universe:
@claude review

think-in-universe:
@ironloopai review

think-in-universe:
@claude review

think-in-universe:
@ironloopai review

think-in-universe:
@ironloopai review

crates/services/src/subscription/service.rs
```diff
@@ -2007,6 +2754,82 @@ fn generate_credit_checkout_idempotency_key(
...
+            return Ok(Some((plan_credits, period_start, period_end)));
+        }
+
+        match self
+            .current_house_of_stake_snapshot_limit(&subscription.subscription_id, Utc::now())
+            .await
+        {
+            Ok(Some((snapshot_limit, snapshot_start, snapshot_end))) => {
+                let snapshot_limit = snapshot_limit.min(i64::MAX as u64) as i64;
+                let reconciliation_plan_credits = plan_credits.max(snapshot_limit);
```
hanakannzashi: **[P1] Do not apply the snapshot floor after a confirmed zero entitlement.**

The chain can authoritatively return zero here when the effective lock has no active shares, but this still replaces that zero with any current snapshot. A user with a 5-credit snapshot and purchased balance is then admitted, and the first 5 credits of post-zero usage are never debited. Preserve a distinct confirmed-zero source, or short-circuit before the snapshot floor, and add coverage for an active snapshot plus a no-active-shares lock and purchased balance.

hanakannzashi (approved):

think-in-universe:
Close in favor of https://github.com/nearai/chat-api/pull/396

PR #371, opened 2026-08-12
feat(migrate): let the operator name a migrated instance on CrabShack

Longarithm:
## Summary

Legacy agent names are **not unique**. Only `instance_id` is `UNIQUE` in `agent_instances`, and the legacy hostname includes the node — so `brave-toad.agent1.near.ai` and `brave-toad.agent3.near.ai` are different agents owned by different people. Measured over all 6,804 records:

| | count |
|---|---|
| names held by more than one owner (non-deleted) | **368** (783 records) |
| names with 2+ **simultaneously active** agents | **109** (229 records) |

`dark-pike` belongs to three unrelated users. Two-word names had no uniqueness guarantee until a 5-char proquint suffix was introduced in 2026-04; the ~3,800 agents created before that are exactly the legacy population still being migrated.

CrabShack's namespace is flat and the name becomes the gateway hostname (`<name>.<zone>`), so only the first agent of a colliding set can keep its name. Today the second one runs the whole migrate path and then fails at import with a 409, which surfaces as a generic 500 `"CrabShack import failed"` — **after** the legacy instance is stopped, so it costs the user a stop/restart cycle and a wasted multi-minute backup, and the error is indistinguishable from the old-version image failure.

This adds an optional `crabshack_name` to `POST /v1/admin/agents/instances/{id}/migrate`. The operator supplies the full new name; on success the instance simply **is** that name — `name` is updated alongside `instance_url`/`dashboard_url`, so there is one name everywhere, the same shape every natively created agent has had since 2026-04. Convention: original name plus a 5-char proquint, `brave-toad` → `brave-toad-abcde`.

**Why the name must be settled before the backup, and why this needs no CrabShack change:** the backup is encrypted to an age recipient derived from the instance name (`derive_age_keypair`), and CrabShack derives the matching identity from the name it is imported under (`instance-import.ts` → `deriveAgeKeypair(passphrase, input.name)`). Deriving from the *target* name up front makes both sides agree, so the fix stays entirely on this side of the wire. It also means the same value must be passed on a retry — a server-generated suffix would differ per attempt and make an existing backup (including one reused via `backup_url`) undecryptable. That is why the operator supplies it rather than the server generating one.

Guards:

- **Reject** anything that is not a single lowercase DNS label. Dots would split the gateway hostname and break log-upload and secret paths.
- A name that does not extend the original is **allowed but logged as a warning** — legal, but usually a copy-paste of the wrong row, and it would strand the agent under a name nobody will look for.

**Uniqueness is the caller's responsibility and is deliberately not checked here** — the operator already queries CrabShack to pick a free name, so re-checking would only duplicate that call on every migrate. A name already in use fails at import with a 409 after the legacy instance is stopped, so the caller pays a stop/restart and a wasted backup for it. That cost is documented on the field.

Operator usage: check `GET $CRAB/instances/<name>`; on 200, try `<name>-<proquint>` until 404, **record the chosen name in the work-list before migrating**, and reuse it on every retry.

## Risk & impact

- **Default path is unchanged.** Omit `crabshack_name` and every derivation, the import name and the URLs use `instance.name` exactly as before, and `name` is not written. The field is `Option`, so existing callers are unaffected.
- **No schema change** and no new dependencies.
- No new outbound calls; the migrate flow is unchanged apart from which name it uses.
- When the field *is* supplied, the user-visible name changes — that is the intent, and those users are already being emailed about the migration changing their URL. The pre-migration name survives in `instance_id` (`agent-<old-name>-<uuid>`, untouched) and in the migrate log's `previous_name=`.
- `admin_update_instance` can now set `name`. Only migration passes it; `admin_patch_instance` passes `None`.
- **Data-classification**: no customer content touched. New log lines carry `instance_id` and system-generated instance names only, matching the existing migrate logging.

## Testing

- `cargo test -p api --lib` → **119 passed**, including 2 new tests in `migrate_name_tests` covering `is_valid_crabshack_name`: the accepted set (incl. proquint-suffixed) and the rejected set (empty, dots, uppercase, underscore, leading/trailing `-`, 64 chars).
- Only the validator is unit-tested. Target-name selection, the prefix warning and the `backup_url` warning live inside the handler and are not covered — the handler is one long function with no seam.
- `cargo clippy --workspace --all-targets` → clean.
- `cargo fmt` applied.
- **No dependency changes** — `Cargo.toml`/`Cargo.lock` untouched, so no new advisories are possible; `cargo-deny` is not installed locally and CI's `security-audit` covers it.
- Not yet exercised end-to-end against a live CrabShack. The prod exercise is a colliding agent from the 109-name set, which is the intended first use.

## Rollback plan

Revert the commits; nothing to undo in the schema. An instance already migrated under a new name keeps working — its `name`, `instance_url` and `dashboard_url` are mutually consistent, and post-revert code addresses it by `name`, which is the name it now has. The only loss is the ability to rename during future migrations.

## Review
- [ ] Security review requested if this is a high-risk change

## Deploy path

Merge to `main` → `.github/workflows/build.yml` builds and publishes on push to `main` (tags `v*` for releases); deployment consumes the published image by digest.

---

Context and measurements — the 368/109 collision counts, the proquint timeline, and the import-409-after-stop analysis — come from a legacy-TEE migration session on 2026-08-12.


oneuron (approved):
okay

PR #372, opened 2026-08-13
refactor(grant-owner): return the owner user id, let the caller grant it

Longarithm:
## Summary

`admin_grant_instance_owner` wrote its grant to CrabShack using the chat-api record's own name:

```rust
"{}/instances/{}/access", crabshack_manager.url, encode(&instance.name)
```

That is not always the name the instance carries on CrabShack. `/migrate` renames the record to the
CrabShack name in its **last** bookkeeping step, so a migrate that failed earlier — or an instance
repaired by hand — keeps the legacy name, which for a row renamed on collision is a **different
tenant's** instance. CrabShack's access route takes any name without checking the instance exists, so
the call returned 200 while writing `role=owner` for this user on the other tenant's agent and leaving
the migrated instance ownerless. Measured on prod 2026-08-13: **11 of 102 renamed instances**.

**The endpoint no longer grants.** It derives the CrabShack user id —
`sha256(hex_decode(auth_secret))`, the one thing only chat-api can compute, since the secret never
leaves it — and returns it. `/migrate` returns the same id alongside the name it imported under.

The caller writes the grant on CrabShack itself, because that is where the decisions live:

- **which name** to grant on — chat-api's record name is not authoritative, the caller's is;
- **whether to revoke** an existing owner row first — CrabShack keys a grant on `(instance, user)`, so
  granting a second user is *additive*: two `role = 'owner'` rows coexist and its owner join then
  returns that instance **twice** in `GET /instances`;
- **whether the instance is far enough along** to own at all.

chat-api could not answer any of those, and guessing is what caused the bug.

Removed along with the write: the CrabShack POST and its error handling, and the "instance must already
point at CrabShack" guard — nothing is written now, and a repair often needs the id *before* the record
has been repointed. Net effect on the file is 93 insertions against 87 deletions.

## ⚠️ Breaking for callers that relied on this granting

The route and method are unchanged, so a caller that ignores the body will now silently stop
establishing owners — and an ownerless instance gets **no scheduled backups** and is invisible to its
user in the portal (CrabShack's watchdog and backup scheduler both filter `status = 'running'` *and*
require an owner). Our migration toolkit is updated in the same breath: it reads `owner_user_id`, then
`POST <crabshack>/instances/<name>/access`, treating an absent id as an older chat-api that granted by
itself. Any other caller needs the same two-step.

Worth deciding in review: the name `grant-owner` now describes what it no longer does. Renaming it
(e.g. `crabshack-owner-id`) would be honest but is a second breaking change, so I left the route alone.

## Risk & impact

- One admin-only endpoint plus one added response field on `/migrate`. No schema, no migration, no
  dependency change (lockfile untouched, so `cargo deny` sees nothing new).
- **Security posture change, flagged deliberately:** the response struct previously documented that it
  *intentionally* withheld this id ("chat-api does not surface credential-derived identity to API
  callers, even admins"). This PR reverses that, because the caller now needs it to grant. The id is a
  one-way hash of the passkey secret and reveals nothing about it, but the decision belongs to a
  CODEOWNER — please confirm rather than rubber-stamp.
- chat-api now makes *fewer* outbound calls: the CrabShack write is gone.

## Testing

- `cargo fmt -p api`, `cargo check -p api` clean; `cargo test -p api --lib` — **123 passing**.
- 3 unit tests on the derivation, including the one that matters: the hash is over the **decoded**
  secret, not its hex text (a wrong owner id would otherwise be silently plausible), plus whitespace
  tolerance and non-hex rejection. Vectors computed independently of the implementation.
- Behaviour verified on prod before the change: grant-owner returned `{"status":"success"}` for renamed
  instances while the CrabShack owner stayed null and the access row appeared on the legacy-named
  tenant instead (`GET /instances/<legacy>/access`).

## Rollback plan

Revert the commits — one file, no state. Callers that already moved to the two-step keep working
against the reverted build, since the old endpoint still granted (on the record name).

## Review

- [ ] Security review — returns a credential-derived identity to admin callers, reversing a documented
      decision; and the bug being fixed granted one user owner access on another user's instance.

## Deploy path

Normal chat-api pipeline. Ship the toolkit change and this together, or landings go ownerless in
between.

## Follow-ups (not in this PR)

1. **CrabShack side:** `POST /instances/<name>/access` should 404 an unknown instance — a grant on a
   name it has never seen currently returns 200, which is what made the bug silent.
2. **Ownership could move into the import.** CrabShack grants an owner at creation
   (`orchestrateCreate` → `grantAccess(db, writer, inst.name, ownerUserId, "owner")`) but
   `orchestrateImport` has no owner concept, so migrated instances are born ownerless and need this
   round trip at all. Passing `owner_user_id` into the import payload would close it.
3. **`/migrate` failure path:** when the post-import DB update fails, chat-api restarts the legacy
   instance and returns an error while the CrabShack copy keeps running under the new name, record
   never repointed — two live copies of one agent.

### DIFF
--- a/crates/api/src/openapi.rs
+++ b/crates/api/src/openapi.rs
@@ -149,7 +149,6 @@ use utoipa::OpenApi;
         services::user::ports::AccountDeletion,
         services::user::ports::AccountDeletionStatus,
         crate::routes::admin::AdminRetryAccountDeletionResponse,
-        crate::routes::admin::GrantInstanceOwnerRequest,
         crate::routes::admin::GrantInstanceOwnerResponse,
         crate::models::AuthResponse,
         crate::error::ApiErrorResponse,
--- a/crates/api/src/openapi.rs
+++ b/crates/api/src/openapi.rs
@@ -149,6 +149,7 @@ use utoipa::OpenApi;
         services::user::ports::AccountDeletion,
         services::user::ports::AccountDeletionStatus,
         crate::routes::admin::AdminRetryAccountDeletionResponse,
+        crate::routes::admin::GrantInstanceOwnerRequest,
         crate::routes::admin::GrantInstanceOwnerResponse,
         crate::models::AuthResponse,
         crate::error::ApiErrorResponse,
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -3647,6 +3647,14 @@ pub struct MigrateInstanceResponse {
     pub status: String,
     pub instance_name: String,
     pub message: String,
+    /// The CrabShack user id to grant as owner, derived from the user's `auth_secret`. A CrabShack
+    /// import lands ownerless, so the caller finishes the job:
+    /// `POST <crabshack>/instances/<instance_name>/access {"user_id": ..., "role": "owner"}` —
+    /// `instance_name` above is the name it was imported under, which is what the grant must use.
+    ///
+    /// `None` when the user has no passkey credentials to derive from, and on the `skipped` path
+    /// (nothing was migrated; use the grant-owner endpoint to look it up).
+    pub owner_user_id: Option<String>,
 }
 
 /// ironclaw-dind tags exist and are allow-listed contiguously from 0.23.0 up.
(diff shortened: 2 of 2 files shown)


oneuron (approved):
pr desc is very confusing but the code is ok)

PR #378, opened 2026-08-19
feat: support AML high-risk policy by risk level or score

### LINKED ISSUE #218: [chat-api] Define high-risk AML wallets by score threshold, not risk level
Parent/context: #35, #55
Related: #89, #202

## Summary

Update `chat-api` AML enforcement so high-risk wallet handling is defined by the provider score, not by the provider `HIGH` risk level string.

The backend should continue to store provider risk level for audit/admin visibility, but the decision for whether a wallet is high-risk should come from an explicit score threshold.

## Context

Current AML tracking and documentation frequently describe high-risk wallets as wallets where Lukka returns `risk_level = HIGH`. Product/compliance policy now needs the high-risk wallet definition to be score-based instead.

This should align Agent Hosting / Private Assistant behavior with the Cloud API policy change, while preserving the existing user-facing privacy boundary: frontend responses must remain neutral and must not expose AML score, provider risk level, report ID, raw provider payload, or provider reason.

## Requirements

- Add or use an explicit `chat-api` AML high-risk score threshold configuration.
- Treat wallets as high-risk when the normalized AML score meets or exceeds that configured threshold.
- Do not rely on provider `risk_level = HIGH` alone to classify or block a wallet.
- Continue persisting provider `risk_level` and `score` in AML reports for audit/admin display.
- Preserve allowlist behavior: an active allowlist entry can override an otherwise high-risk score.
- Preserve neutral public responses and frontend-facing error behavior.
- Make Slack alerts, metrics, and logs use the score-threshold decision when reporting a high-risk block.
- Keep source-app identification from #202 compatible with this change.
- Define behavior for missing/non-numeric scores explicitly, likely treating the result as `UNKNOWN`/provider failure rather than high-risk-by-level.

## Tests

Add or update tests for:

- A report with score at or above the threshold is treated as high-risk even if provider `risk_level` is not `HIGH`.
- A report with score below the threshold is not treated as high-risk solely because provider `risk_level` is `HIGH`.
- Missing or malformed score follows the configured unknown/provider-failure path.
- Allowlisted accounts can proceed even when their score meets the high-risk threshold.
- Stored/admin report fields still include both `risk_level` and `score`.
- Slack alert tests assert the score-threshold high-risk decision without leaking secrets or raw provider payloads.
- Public/frontend responses do not expose score, risk level, provider reason, or report details.

## Acceptance criteria

- `chat-api` defines high-risk wallet handling by configured AML score threshold.
- Provider risk level remains audit metadata, not the primary high-risk predicate.
- Existing Agent Hosting / Private Assistant NEAR wallet flows use the score-threshold decision.
- Tests cover above-threshold, below-threshold, allowlisted, Slack alert, and missing-score behavior.

think-in-universe:
Fixes nearai/house-of-stake-contracts#218.

## Summary
- add configurable AML high-risk risk-level policy (`LUKKA_AML_HIGH_RISK_LEVELS`, alias `LUKKA_AML_BLOCKED_RISK_LEVELS`)
- make AML score-threshold policy configurable and optional (`LUKKA_AML_HIGH_RISK_SCORE_THRESHOLD`, alias `LUKKA_AML_SCORE_BLOCK_THRESHOLD`)
- classify high-risk results deterministically when either configured predicate matches
- preserve risk-level-only screening when score is missing or malformed, while threshold-only policy treats missing/malformed scores as provider failure
- apply the configured policy to user status, HoS subscription AML enforcement, stale fallback, logs, and Slack alerts
- keep admin/audit risk level and score fields while removing provider AML details from public responses

## Validation
- cargo fmt --check
- cargo check -q
- cargo test -q -p config
- cargo test -q -p services aml
- cargo test -q -p api routes::subscriptions::tests
- cargo test -q -p api --test subscriptions_tests test_change_plan_outcome_serde_uses_kind_discriminant

## Not Run
- cargo test -q -p api --test subscriptions_tests test_resume_subscription_house_of_stake: local Postgres rejected the configured postgres password during migrations


crates/services/src/aml.rs
```diff
@@ -223,6 +276,18 @@ pub trait AmlRiskService: Send + Sync {
     fn alert_on_cached_reports(&self) -> bool {
         false
     }
+    fn high_risk_risk_levels(&self) -> Vec<AmlRiskLevel> {
+        vec![AmlRiskLevel::High]
+    }
+    fn high_risk_score_threshold(&self) -> Option<i64> {
+        Some(DEFAULT_AML_HIGH_RISK_SCORE_THRESHOLD)
+    }
+    fn is_high_risk_result(&self, result: &AmlCheckResult) -> bool {
```
hanakannzashi: [P1] Once this policy supports score-only enforcement, fresh reports produced before this rollout are not qualified against the current policy. A pre-rollout active `HIGH` report with `score = NULL` remains retrievable; with risk levels disabled this predicate returns false, and the subscription/user services return that report before calling Lukka. The account can be allowed for the full refresh window even if a fresh score would block it. Treat cached score-less reports as unknown and refresh them in score-only mode.

crates/services/src/aml.rs
```diff
@@ -653,39 +748,36 @@ enum LukkaAuthMode<'a> {
...
+
+    if require_score && score.is_none() {
+        return unknown_lukka_response(account_id, report, cscore, "provider_missing_score");
+    }
+
+    let provider_risk_level = cscore
+        .as_ref()
+        .and_then(|section| section.risk_level.as_deref());
+    let risk_level =
+        AmlRiskLevel::from_provider(provider_risk_level).unwrap_or(AmlRiskLevel::Unknown);
```
hanakannzashi: [P1] A score-only decision with a missing or unrecognized risk level can be correctly blocked on its first check because the score is usable, but it is not durable: `record_report` marks every `UNKNOWN` result inactive, `latest_active_report` excludes it, and `store_cache` skips this non-provider `UNKNOWN`. A later provider outage therefore has no stale high-score fallback and fails open. Make report/cache eligibility policy-aware rather than `risk_level != UNKNOWN`.

crates/services/src/aml.rs
```diff
@@ -653,39 +748,36 @@ enum LukkaAuthMode<'a> {
...
+    let reason = if risk_level == AmlRiskLevel::Unknown {
+        Some(
+            provider_risk_level
+                .map(|value| format!("unrecognized_risk_level:{}", value.trim()))
```
hanakannzashi: [P1] This makes an absent or unrecognized provider risk level a non-`provider_` result. In a risk-level/combined deployment with no usable score, `is_provider_failure()` is then false, so the subscription and user flows skip their stale-HIGH fallback and allow the account. The previous implementation classified this as `provider_invalid_response`; retain an anomaly/provider-failure path whenever a configured risk-level predicate has no usable signal.

hanakannzashi (commented):
AML policy review: three high-risk enforcement gaps are noted inline.

env.example
```diff
@@ -125,6 +125,8 @@ STRIPE_WEBHOOK_SECRET=
 LUKKA_AML_ENABLED=false
 LUKKA_BASE_URL=https://api.blockchain-analytics.lukka.tech
 LUKKA_BEARER_TOKEN=
+LUKKA_AML_HIGH_RISK_LEVELS=HIGH
+LUKKA_AML_HIGH_RISK_SCORE_THRESHOLD=75
```
think-in-universe: We'd prefer the env var `LUKKA_AML_SCORE_BLOCK_THRESHOLD`

crates/api/src/routes/subscriptions.rs
```diff
@@ -89,14 +89,12 @@ impl From<CreateSubscriptionOutcome> for CreateSubscriptionResponse {
 
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
 pub struct PublicAmlCheckResult {
-    pub risk_level: AmlRiskLevel,
```
think-in-universe: Why remove the `risk_level`? We should still return the risk level from Lukka.

crates/services/src/aml.rs
```diff
@@ -69,8 +71,12 @@ impl AmlCheckResult {
         }
     }
 
+    pub fn is_high_risk_at_score_threshold(&self, threshold: i64) -> bool {
+        self.score.is_some_and(|score| score >= threshold)
+    }
+
     pub fn is_high_risk(&self) -> bool {
-        self.risk_level == AmlRiskLevel::High
+        self.is_high_risk_at_score_threshold(DEFAULT_AML_HIGH_RISK_SCORE_THRESHOLD)
```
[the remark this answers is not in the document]
think-in-universe: Addressed in bba58c1 by removing the hardcoded public AML redaction policy. Successful responses now expose only the public-safe AML metadata (`risk_level` and `checked_at`), while blocked AML flows return before response serialization.

crates/services/src/subscription/service.rs
```diff
@@ -350,12 +351,15 @@ impl SubscriptionServiceImpl {
...
-        if result.is_high_risk() {
+        if self.aml_service.is_high_risk_result(result) {
             tracing::warn!(
                 user_id = %user_id,
                 flow = %flow,
                 report_id = ?result.report_id,
-                "High-risk AML result for NEAR wallet flow"
+                score = ?result.score,
+                high_risk_risk_levels = ?self.aml_service.high_risk_risk_levels(),
+                high_risk_score_threshold = self.aml_service.high_risk_score_threshold(),
```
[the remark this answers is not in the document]
think-in-universe: Addressed in bba58c1. The production `warn!` in HoS enforcement no longer logs the AML score; it keeps user/flow/report id plus configured policy metadata.

crates/services/src/aml.rs
```diff
@@ -589,15 +676,23 @@ impl LukkaAmlService {
...
                                 "Lukka AML check completed"
                             );
-                            if result.is_high_risk() {
+                            if self.is_high_risk_result(&result) {
                                 tracing::warn!(
                                     report_id = ?result.report_id,
-                                    "Lukka AML high risk NEAR account detected"
+                                    score = ?result.score,
+                                    high_risk_risk_levels = ?self.config.high_risk_risk_levels,
+                                    high_risk_score_threshold = self.config.high_risk_score_threshold,
```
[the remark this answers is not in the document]
think-in-universe: Addressed in bba58c1. Removed the AML score from the Lukka high-risk production warning log. I also added a missing-score warning for configured score policy without logging the score value.

crates/services/src/subscription/service.rs
```diff
@@ -317,10 +317,11 @@ impl SubscriptionServiceImpl {
         self.alert_aml_provider_failure(user_id, flow, &result);
         if result.is_provider_failure() {
             // Compliance decision (2026-07-18): Lukka provider outages fail open for
-            // user-initiated HoS billing flows unless a stale active HIGH report is
+            // user-initiated HoS billing flows unless a stale active high-score report is
             // already known for the NEAR account. Provider failures are recorded and
             // alerted so compliance can review outage-period activity.
```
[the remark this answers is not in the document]
think-in-universe: Addressed in bba58c1. The stale fallback comment now refers to the configured high-risk policy instead of a high-score-only report.

crates/api/src/routes/subscriptions.rs
```diff
@@ -89,14 +89,12 @@ impl From<CreateSubscriptionOutcome> for CreateSubscriptionResponse {
...
     pub checked_at: DateTime<Utc>,
 }
```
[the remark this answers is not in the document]
think-in-universe: Addressed in bba58c1. `public_aml_result` no longer uses `AmlCheckResult::is_high_risk()` or any hardcoded policy. Successful responses return a public-safe value only (`risk_level`, `checked_at`); blocked high-risk flows do not serialize these responses.

crates/services/src/aml.rs
think-in-universe: Addressed in bba58c1. Added `has_usable_policy_signal` and made the subscription/user services ignore fresh cached active reports that do not qualify under the current policy, so score-only mode refreshes pre-rollout scoreless reports instead of allowing them for the refresh window.

crates/services/src/aml.rs
think-in-universe: Addressed in bba58c1. Report/cache eligibility is now policy-aware: `AmlReportEvent` carries the active decision, `latest_active_report` no longer excludes `UNKNOWN`, cache storage uses the same policy-aware active decision, and active UNKNOWN reports bypass the UNKNOWN dedupe so score-only high-score results stay durable.

crates/services/src/aml.rs
think-in-universe: Addressed in bba58c1. Missing or unrecognized Lukka `risk_level` now uses `provider_missing_risk_level` / `provider_unrecognized_risk_level:*`, so it is treated as a provider failure for alerting, negative caching, and stale fallback behavior.

env.example
think-in-universe: Addressed in bba58c1. `env.example` now documents `LUKKA_AML_SCORE_BLOCK_THRESHOLD`, and the config parser prefers that name while still accepting the previous alias for compatibility.

crates/api/src/routes/subscriptions.rs
think-in-universe: Addressed in bba58c1. Restored `risk_level` in `PublicAmlCheckResult`; public AML responses now include `risk_level` and `checked_at`, while still omitting score, report id, provider, account id, address type, and reason.

think-in-universe:
@claude Addressed in bba58c1:

- Score blocking is now opt-in; absent score env means risk-level-only behavior, and `AmlCheckResult::is_high_risk()` no longer hardcodes score >= 75.
- Missing/unrecognized Lukka risk levels are provider anomalies again (`provider_missing_risk_level` / `provider_unrecognized_risk_level:*`) for alerting, negative caching, and stale fallback.
- `require_score` is derived from parsed risk levels, and a configured but unusable score emits a warning without logging the score value.
- Public AML responses no longer use hardcoded high-risk filtering. Successful HoS responses include public-safe `risk_level` + `checked_at`; blocked flows return before serialization.
- Explicit malformed policy config no longer silently falls back to stricter blocking. Score threshold typos/out-of-range values become disabled, and risk-level typos become an empty configured list; absent risk levels still default to `HIGH` for existing behavior.
- Report/cache eligibility is policy-aware so score-only UNKNOWN/high-score reports remain durable, while scoreless pre-rollout reports are refreshed.
- Removed AML score from production warn logs and updated the stale fallback comment.

Validation run locally: `cargo test -p config lukka_aml`, `cargo test -p services aml::tests`, `cargo test -p api public_aml`, `cargo test -p database aml`, `cargo fmt --check`, and `cargo check --workspace`. The DB-backed subscription integration tests compile but cannot run in this environment because local PostgreSQL auth fails for user `postgres` during migrations.

think-in-universe:
@ironloopai review

env.example
```diff
@@ -125,6 +125,8 @@ STRIPE_WEBHOOK_SECRET=
 LUKKA_AML_ENABLED=false
 LUKKA_BASE_URL=https://api.blockchain-analytics.lukka.tech
 LUKKA_BEARER_TOKEN=
+LUKKA_AML_HIGH_RISK_LEVELS=HIGH
+LUKKA_AML_SCORE_BLOCK_THRESHOLD=75
```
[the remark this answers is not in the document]
think-in-universe: Thanks for catching this. The only Compose file in the repo is `docker-compose.yml`, and that deployment path is not currently in use for this service, so we are intentionally leaving the Compose env mapping unchanged for now. The active deployment path consumes the documented env vars directly.

think-in-universe:
@claude review

crates/database/src/repositories/aml_report_repository.rs
```diff
@@ -153,7 +152,7 @@ impl AmlReportRepository for PostgresAmlReportRepository {
                 r#"
                 SELECT *
                 FROM aml_risk_reports
-                WHERE account_id = $1 AND active = TRUE AND risk_level <> 'UNKNOWN'
+                WHERE account_id = $1 AND active = TRUE
```
hanakannzashi: [P1] This `LIMIT 1` runs before policy usability is evaluated. Existing rows can contain an older active `LOW/score=90` report and a newer active `LOW/score=NULL` report, because pre-rollout code marked every known risk level active. After enabling a score threshold, the newer row is unusable; if Lukka then times out, the stale fallback examines only that row, fails open, and never sees the older score match. Fetch/filter enough candidates (or expose separate selectors for fresh usability and stale policy-high-risk fallback) instead of using one newest row for both, and add a two-row provider-timeout regression test.

hanakannzashi: [P1] Pre-score-policy releases preserved numeric scores on `UNKNOWN` provider responses but persisted every such report with `active = false`. This change now treats `UNKNOWN + score >= threshold` as high risk and records new rows active, but there is no migration/reclassification for existing rows, so this predicate can never return them. A later provider outage therefore fails open despite a legacy score that matches the new policy. Add a migration/provenance strategy and a seeded legacy-row outage regression test; do not simply drop `active`, since admins may have intentionally deactivated reports.

crates/api/src/main.rs
```diff
@@ -36,6 +36,85 @@ fn near_rpc_network_hint(host: Option<&str>) -> Option<&'static str> {
...
+            }
+        }
+    }
+
+    if config.enabled
+        && config.high_risk_risk_levels.is_empty()
+        && config.high_risk_score_threshold.is_none()
+    {
+        tracing::warn!(
+            "Lukka AML is enabled but both high-risk predicates are disabled; AML checks will not block accounts"
```
hanakannzashi: [P1] This only warns and continues after parsing invalid values into zero active predicates. For example, with `LUKKA_AML_ENABLED=true`, `LUKKA_AML_HIGH_RISK_LEVELS=HGIH`, and the score threshold disabled, config becomes `[]/None` and every AML result is allowed. Reject invalid values and a zero-predicate policy at startup when AML is enabled, or require a separate explicit observe-only mode.

think-in-universe:
@ironloopai resolve

Please help resolve the review comments properly

hanakannzashi (approved):

PR #382, opened 2026-08-20
feat(api): Stage I owner-only views and stateless Responses

hanakannzashi:
Closes #386
Closes #387
Closes #388
Closes #389

## Summary

This PR implements the Stage I owner-only read contract in #385. It preserves the existing authenticated owner views needed for migration/export, disables ordinary Conversation/File and all sharing routes, and keeps the Responses proxy stateless.

## Stage I API contract

This table covers the Private Chat stateful surface only. Other inference APIs remain outside this migration scope.

| Surface | Method and path | Auth | Stage I treatment |
| --- | --- | --- | --- |
| Stateless Responses | `POST /v1/responses` | Existing dual-auth, subscription, and rate-limit boundary | **Retained.** Every response, including local middleware and validation errors, is `Cache-Control: no-store`. A non-empty body must be an identity-encoded JSON object. Chat rejects `store: true`, non-null `conversation`, non-null `previous_response_id`, and `background: true`; it normalizes valid JSON requests to `store: false`. It does not inject author metadata, perform Conversation ACL/tracking, execute tools, or run an agent loop. |
| Client-managed functions | `tools: [{type:"function"}]` and replayed `input` items | Same as Responses | **Retained.** After existing request normalization, Chat does not turn these items into another transcript shape, reorder them, or execute them. The caller supplies prior context and matching function-call output on the next request. |
| Responses built-ins / remote MCP | Built-in or remote-MCP tool/input shapes submitted to `/v1/responses` | Same as Responses | **Not implemented by Chat.** Chat forwards the shape to Cloud; Cloud applies its stateless Responses validation and returns its normal `400` for unsupported capabilities. This does not affect the separate root MCP proxy below. |
| Owner Conversation views | `GET /v1/conversations`<br>`GET /v1/conversations/{conversation_id}`<br>`GET /v1/conversations/{conversation_id}/items` | Session auth; current user must own the record | **Temporarily retained** for migration/export. All responses are `Cache-Control: no-store`. |
| Owner File views | `GET /v1/files`<br>`GET /v1/files/{file_id}`<br>`GET /v1/files/{file_id}/content` | Session auth; existing caller/ownership checks | **Temporarily retained** for migration/export. All responses are `Cache-Control: no-store`. |
[...]

PR #383, opened 2026-08-20
[Decision][Epic 953][Stage I][chat-api] Preserve account deletion across Cloud migration

### LINKED ISSUE #380: Stop account-deletion worker from calling retired Cloud Conversations and Files APIs
Parent: #377

## Goal

Keep account deletion reliable after Cloud Conversations and Files endpoints are retired. The current worker treats a retired upstream endpoint as a retryable error, then eventually marks deletion failed_needs_review.

## Scope

- Remove the task-worker dependency on Cloud conversation/file deletion calls.
- Update deletion progress so it no longer requires cloud_deleted_conversation_ids or cloud_deleted_file_ids before finalization.
- Preserve safe handling of existing in-flight progress records; old progress fields may be ignored but must not make a deletion fail.
- Remove now-unused provider-delete helper methods and their focused tests only when no other worker path uses them.
- Add regression coverage for deleting an account with legacy local conversation/file rows when Cloud returns 410 or is not contacted.

## Non-goals

- Do not drop or migrate Conversation/File/share tables.
- Do not perform bulk data deletion, retention changes, or external object cleanup beyond the normal existing account-deletion flow.
- Do not change the public route behavior; that belongs to #379.

## Acceptance criteria

- Account deletion completes without a Cloud Conversations or Files request.
- A Cloud 410 cannot cause retrying or failed_needs_review for this workflow.
- Legacy account-deletion progress remains readable and safe to resume.

## Dependencies

- Parent: #377
- Coordinate rollout with #379 and nearai/cloud-api#934.

hanakannzashi:
Tracks #380

## Resolved — no Chat API code change required

The existing Chat account-deletion contract remains unchanged in Stage I:

- `DELETE /v1/users/me` remains the authenticated asynchronous account-deletion endpoint.
- Accepted jobs keep the established worker behavior, including Cloud Conversation/File cleanup followed by local finalization.
- Cloud #943 retains `DELETE /v1/conversations/{conversation_id}` and `DELETE /v1/files/{file_id}` as the explicit Stage I compatibility exception, so the worker continues to use its existing API-key/workspace-scoped calls.
- This is a retained public resource-delete surface, not a new trusted Chat-only token or internal Cloud path. Other Cloud Conversation/File mutations remain retired.

## Why this PR is closed

The earlier worker-change direction—removing Cloud cleanup calls while retaining local finalization—would have produced partial deletion and is intentionally not used. With the retained Cloud DELETE endpoints, no Chat worker or user-facing API change is required for the selected Stage I plan.

This closed PR is retained as the decision record. Account-deletion verification is tracked by #380 and must be included in the Cloud/Chat rollout.

## Out of scope

- Changing or disabling the user-facing account-deletion contract.
- Removing worker cleanup or local finalization.
- Frontend changes, archive/tombstone work, or Stage III cleanup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ A Rust backend service that proxies OpenAI-compatible inference requests to **NE
 ```
 crates/
 ├── api/          # Axum HTTP server, routes, middleware, OpenAPI docs (utoipa)
-├── services/     # Business logic: auth, response proxy, user management
+├── services/     # Business logic: auth, temporary read views, response proxy, user management
 ├── database/     # PostgreSQL (tokio-postgres, deadpool), migrations, repositories
 └── config/       # Environment-based configuration structs
 ```
@@ -30,13 +30,15 @@ crates/
 - **Repository Pattern**: Database access through trait-based repositories (`PostgresUserRepository`, etc.)
 - **Service Layer**: Business logic in `services` crate, injected into `AppState`
 - **NEAR AI Cloud API Proxy**: OpenAI-compatible inference routes forward to NEAR AI Cloud API with auth; Responses requests are stateless
+- **Temporary Read Views**: Owner-only Conversation and File GET endpoints remain available for the Stage I migration/export window. Sharing surfaces and all write operations return `410 Gone`.
 - **Patroni Support**: Optional cluster discovery for HA PostgreSQL via `DATABASE_PRIMARY_APP_ID`
 
 ### Request Flow
 
-1. Request → Auth middleware (validates session token) → Route handler
-2. `/v1/responses` → validate stateless fields → forward to NEAR AI Cloud API with `store: false`
-3. Usage, subscription, rate-limit, and attestation-related proxy behavior remain local to Chat API
+1. Request → its historical authentication boundary → route handler
+2. `/v1/responses` → validate stateless linkage fields → forward to NEAR AI Cloud API with `store: false`
+3. Temporary owner-only Conversation/File GET views → local ownership lookup and, where needed, Cloud read view
+4. Usage, subscription, rate-limit, and attestation-related proxy behavior remain local to Chat API
 
 ## Development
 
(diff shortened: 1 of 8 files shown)


hanakannzashi:
## Stage I rollout clarification

Cloud API's Conversation and File mutation endpoints are write-disabled in Stage I. This worker change is therefore a Stage I compatibility prerequisite, not final stateful-surface removal: account deletion must not retry those upstream deletes or fail because they return `410 Gone`.

It preserves local account-deletion finalization and does not alter the temporary read/export views. Deploy it before or with the Stage I proxy boundary and Cloud API #943; final removal remains Stage III work.

hanakannzashi:
Update: the Stage I decision is now to disable self-service account deletion, not to preserve local finalization behind an archive/deferred choice. This Draft must be reworked to return `410 Gone` before task creation and safely recover existing in-flight deletion requests; it must not delete the ownership manifest.

hanakannzashi:
Clarification: the Stage I decision is not to cancel accepted account-deletion jobs. New requests become `410` before task creation; pre-cutoff jobs keep their legacy cleanup path and must drain before Cloud disables its delete endpoints. This supersedes the prior Draft comment about recovering existing jobs.

hanakannzashi:
Reworked at `62c5d71` on the current #382 base. New `DELETE /v1/users/me` (including `/me/`) returns authenticated `410 Gone` with `Cache-Control: no-store` before any deletion row or task is created. The legacy worker, Cloud/local cleanup path, and admin retry remain unchanged so pre-cutoff jobs can drain normally. OpenAPI and regression tests now cover the cutoff; CI has been requeued.

hanakannzashi:
Final Stage I decision: normal `DELETE /v1/users/me` and its existing worker remain active; Cloud #943 retains both resource DELETE routes. This supersedes the earlier cutoff/410 and worker-removal comments above. No code from this closed PR is to be merged; #380 owns end-to-end rollout verification.

Issue #381, opened 2026-08-20
Remove dormant Private Chat services and runtime wiring


hanakannzashi:
Parent: #377

## Goal

After HTTP retirement and account-deletion compatibility land, remove the application/runtime code that only supported Private Chat Conversations, Files, sharing, and stateful Responses behavior.

## Scope

- Remove unused Conversation, ConversationShare, and File service fields from AppState and test app setup.
- Remove their construction from API startup and any task-worker setup after the worker no longer needs them.
- Remove unused provider proxy helpers, repositories, ports, metrics/analytics writes, and test fixtures that only support retired Private Chat behavior.
- Remove obsolete README/OpenAPI/internal architecture references left after #379.
- Keep database schema and historical rows intact.

## Non-goals

- No schema migrations, table drops, retention/export work, or external object deletion.
- Do not remove supported generic proxy behavior, usage accounting, subscriptions, rate limits, model settings, MCP, or attestation.
- Do not change public endpoint semantics; #379 owns that boundary.

## Acceptance criteria

- API startup and tests no longer instantiate unused Private Chat services.
- No production request/worker path holds a provider-backed Conversation/File service.
- No schema/data changes are included.
- Supported proxy tests remain green.

## Dependencies

- Parent: #377
- Depends on #379 and the account-deletion worker child issue created alongside it.

hanakannzashi:
## Stage III deferral clarification

Earlier all-`410` planning made this cleanup appear immediately safe. Under the approved Stage I plan, Conversation/File/Share services and repositories remain live behind temporary authenticated read/export views.

Do not remove that wiring during Stage I. This issue begins only after the export window closes and #390–#392 remove the final stateful route surface.

PR #384, opened 2026-08-20
[Draft][Epic 953][Stage III][chat-api] Deferred runtime cleanup after the export window

hanakannzashi:
Tracks #381.

## Draft — Stage III only

**The current head is historical only.** It was built for the superseded all-`410` and account-deletion-removal design. It must not be reviewed, rebased, or merged during Stage I.

This PR is retained as a Draft placeholder only. It is not part of the active #382 Stage I rollout stack, and its current diff is not authoritative for the current contract.

## Future scope

After the migration/export window closes, #390, #391, and #392 remove the final stateful proxy routes, and an approved account-deletion compatibility plan is in place, rebuild this work from then-current `main` for #381. Revalidate all code, docs, tests, and runtime dependencies at that time.

## Explicitly not current scope

- Removing Stage I authenticated owner read views or their backing Conversation/File service/repository wiring.
- Changing the normal `DELETE /v1/users/me` account-deletion workflow.
- Schema/table deletion, retention purge, external-object cleanup, or new export behavior.
- Changing the stateless Responses contract.


hanakannzashi:
## Stage III draft clarification

This draft was built on the superseded all-`410` approach. Its current removal of Conversation/File/Share runtime wiring is incompatible with Stage I because the temporary authenticated read/export views still use that wiring.

It is intentionally deferred and no longer part of the active Stage I stack. Rework/rebase it only after the Stage III route-removal issues (#390–#392) are complete.

hanakannzashi:
Supersedes earlier review approvals: they applied to the superseded stack, not the current contract. Do not merge, rebase, or use this diff for Stage I. After #390–#392 and confirmation of an approved account-deletion path, rebuild from then-current `main` and revalidate before using this PR.

hanakannzashi:
Closing this historical Stage III draft. Its current diff was built for the superseded all-410/account-deletion-removal design and must not be reused. After the export/migration window closes, the replacement path is live and verified, and account-deletion compatibility is resolved, recreate the runtime-cleanup implementation from then-current `main` under #381.

PR #395, opened 2026-08-23
fix: HoS accounting review comments

think-in-universe:
## Summary

This stacked PR addresses the latest HoS accounting review comments on #363.

- Distinguish persisted-snapshot fallback from plan-only fallback, and block over-plan proxy usage when no authoritative entitlement snapshot is available.
- Bypass the 10-minute HoS entitlement cache for purchased-credit reconciliation so direct on-chain stake increases cannot be charged as stale overage.
- Persist confirmed zero-stake observations to reset the carried stake baseline.
- On a fixed-to-stake-based plan change, retain the fixed current-period entitlement and prorate only the variable-plan increase.
- Add regression coverage for confirmed zero stake, direct chain stake increases, and fixed-to-variable proration.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo check --all-targets --features test`
- `cargo test -p database purchased_reconciliation`

## Base

Stacked on #363 (`agent/prorate-hos-stake-credits`).

### DIFF
--- a/crates/api/tests/subscriptions_tests.rs
+++ b/crates/api/tests/subscriptions_tests.rs
@@ -5900,7 +5900,6 @@ async fn test_house_of_stake_purchased_reconciliation_uses_synced_credit_limit_c
         let mut state = state.lock().expect("lock HoS mock state");
         state.amount_near = yocto_near(1000);
     }
-    sync_house_of_stake_subscription(&server, &token).await;
 
     let user = db
         .user_repository()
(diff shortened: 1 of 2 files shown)


think-in-universe (commented):
Reviewed for correctness, security, performance/concurrency, tests, and repository conventions.

Found one high-confidence performance issue: purchased-credit reconciliation now performs uncached NEAR RPC reads in the per-usage request path. The accounting behavior is otherwise consistent with the stated fixes and regression coverage.

hanakannzashi (approved):

PR #396, opened 2026-08-23
feat: prorate HoS stake credit entitlement

### LINKED ISSUE #132: Prorate stake-based monthly credits for mid-period stake increases

think-in-universe:
## Summary

Replaces the stacked implementation in #363 and #395 with one commit based directly on `main`. The implementation is intentionally limited to the accounting state needed for proration and the cache behavior needed to keep NEAR RPC off the request hot path.

Final scope: **5 files, 504 additions, 302 deletions (806 changed lines)**.

### Changes

- Add a compact V37 table keyed by HoS subscription and local billing period. It stores the credited stake high-water mark, last observed stake, granted plan credits, and observation time.
- Use one entitlement transition: the period baseline receives full credits; a later observed stake increase receives only the fraction remaining in the period.
- Carry the last observed stake into the next period. Decreases do not claw back credits already granted in the current period; a confirmed zero stake returns zero entitlement and records zero for the next baseline.
- Make the existing 10-minute credit-limit cache part of the shared plan resolver used by proxy checks, `/v1/credits`, and post-usage purchased-credit reconciliation. Reconciliation no longer invalidates the plan cache.
- On a transient chain lookup failure, use only the current persisted snapshot. Without a snapshot, fail closed and negative-cache the RPC failure for 60 seconds.
- Keep the purchased-credit reconciliation cursor as a same-period high-water mark so a later plan-limit increase cannot charge the same overage twice.
- Extend the existing HoS integration helper to read credits twice and assert that the second read adds no NEAR RPC calls.

### Design assumptions

- `subscriptions.current_period_end` is the billing-period boundary; the start is derived with the existing one-month helper. This change does not introduce a second chain-period model.
- The contract exposes current effective stake, not the stake-change timestamp. Proration starts when the backend first observes an increase.
- The first-ever snapshot uses the observed stake as the full-period baseline to avoid reducing existing users during rollout. Later periods use the previous snapshot's last observed stake as their baseline.
- Direct chain changes become visible after explicit HoS sync or the normal 10-minute cache expiry. No per-request RPC refresh is attempted.
- Stake decreases do not claw back current-period entitlement. Confirmed inactive/zero-share locks return zero immediately and persist a zero observation.
- A transient RPC failure may use a snapshot for the exact local period only. If none exists, access fails closed for the 60-second failure-cache window instead of trusting a configured static limit.
- Subscription renewal and price changes update the local subscription row through the existing sync/webhook paths. This PR does not add special in-period fixed-to-variable migration semantics.

## Risk & impact

This changes plan-credit and purchased-balance accounting for stake-based House-of-Stake subscriptions. The migration is additive. Stripe and free-plan accounting keep their existing behavior, but now share the same resolver cache already used by proxy checks. RPC load is bounded by the 10-minute successful-result cache and 60-second failure cache. No customer content or credentials are logged.

## Testing

- `cargo fmt --all -- --check`
- `cargo check --all-targets --features test`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test -p database repositories::credits_repository::tests` (2 passed)
- `cargo test -p services subscription::service::tests` (25 passed)
- `cargo test --test subscriptions_tests --features test` (88 passed)
- `cargo deny check advisories bans sources` reports advisories already present in the current lockfile. This PR does not change `Cargo.lock` or dependency versions; the repository workflow records the audit as report-only.

## Rollback plan

Revert the single commit. The additive snapshot table can remain because the previous code does not read it; remove it later with a separate migration if needed.

## Review

- [ ] Security review requested if this is a high-risk change

## Deploy path

Merge to `main` and deploy through the standard immutable image pipeline.

### DIFF
--- a/crates/api/tests/subscriptions_tests.rs
+++ b/crates/api/tests/subscriptions_tests.rs
@@ -4042,21 +4042,24 @@ async fn assert_house_of_stake_credits_for_effective_lock(
     insert_house_of_stake_subscription_for_existing_user(&db, near_email, "price_hos_basic", false)
         .await;
 
-    let response = server
-        .get("/v1/credits")
-        .add_header(
-            http::HeaderName::from_static("authorization"),
-            http::HeaderValue::from_str(&format!("Bearer {token}")).unwrap(),
-        )
-        .await;
+    for _ in 0..2 {
+        let response = server
+            .get("/v1/credits")
+            .add_header(
+                http::HeaderName::from_static("authorization"),
+                http::HeaderValue::from_str(&format!("Bearer {token}")).unwrap(),
+            )
+            .await;
 
-    assert_eq!(response.status_code(), 200, "{}", response.text());
-    let body: serde_json::Value = response.json();
-    assert_eq!(
-        body.get("plan_credits").and_then(|x| x.as_i64()),
-        Some(expected_plan_credits),
-        "{assertion_message}"
-    );
+        assert_eq!(response.status_code(), 200, "{}", response.text());
+        let body: serde_json::Value = response.json();
+        assert_eq!(
+            body.get("plan_credits").and_then(|x| x.as_i64()),
+            Some(expected_plan_credits),
+            "{assertion_message}"
+        );
+    }
+    assert_eq!(mock.received_requests().await.unwrap().len(), 2);
 }
 
 #[tokio::test]
--- a/crates/api/tests/subscriptions_tests.rs
+++ b/crates/api/tests/subscriptions_tests.rs
@@ -4131,6 +4131,115 @@ async fn test_house_of_stake_credits_zero_for_effective_lock_with_no_active_shar
     .await;
 }
 
+#[tokio::test]
+#[serial(subscription_tests)]
+async fn test_fixed_house_of_stake_credits_persist_authoritative_period() {
+    clear_proxy_env_for_local_wiremock();
+    let now = Utc::now().with_nanosecond(0).unwrap();
+    let period_start = now - Duration::days(20);
+    let period_end = now + Duration::days(10);
+    let subscription_id = format!("sub_chain_hos_fixed_{}", Uuid::new_v4());
+    let chain_sub = json!({
+        "subscription_id": subscription_id,
+        "price_id": "price_hos_starter",
+        "last_lock_id": "lock_chain_hos_fixed",
+        "start_ns": period_start.timestamp_nanos_opt().unwrap().to_string(),
+        "end_ns": period_end.timestamp_nanos_opt().unwrap().to_string(),
+        "status": "Active",
+        "cancel_at_period_end": false
+    });
+    let mock = MockServer::start().await;
+    Mock::given(method("POST"))
+        .and(path("/"))
+        .respond_with(near_rpc_wiremock_hos_subscription_by_price(
+            "price_hos_starter",
+            chain_sub,
+            None,
+        ))
+        .mount(&mock)
+        .await;
+
+    let (server, db) = create_test_server_and_db(TestServerConfig {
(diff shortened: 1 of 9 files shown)


think-in-universe (commented):
Reviewed against PR #363 behavior and the existing inline findings. I found two additional correctness gaps. The first is a feature-parity blocker for the upgrade use case this PR is intended to replace; the second can serve stale entitlement across a billing boundary. Existing CI is green, but the current tests do not cover either case.

hanakannzashi (approved):

